PR #2, opened 2024-01-04
MAP-524 Initial setup of the API

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/config/WebClientConfiguration.kt
```diff
@@ -0,0 +1,66 @@
...
+
+@Configuration
+class WebClientConfiguration(
+  @Value("\${api.base.url.oauth}") private val authBaseUri: String,
+) {
+
+  @Bean
+  fun webClient(builder: WebClient.Builder): WebClient {
+    return builder
+      .baseUrl(authBaseUri)
```
petergphillips, line 30: worth adding timeout to this too

src/main/resources/application-localstack.yml
```diff
@@ -0,0 +1,14 @@
+hmpps.sqs:
+  reactiveApi: true
```
petergphillips, line 2: not sure if you mean this to use the reactive api

petergphillips (approved):
LGTM

src/main/resources/application-localstack.yml
Mjwillis: no definitely not!

PR #12, opened 2024-01-12
MAP-580

Mjwillis:
- Provide a simple update and delete endpoint
 - Remove need to supply hierarchy and allow dynamic re-calc of location path dependant on parent location


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -56,11 +52,49 @@ class LocationService(
...
+        "path" to locationToUpdate.getPathHierarchy(),
+      ),
+      null,
+    )
+
+    return locationToUpdate.toDto()
+  }
+
+  @Transactional
+  fun deleteLocation(id: UUID): LocationDTO {
```
andymarke, line 92: delete is interesting, since deleting in NOMIS can cause FK error - I wonder how we will deal with this

andymarke (approved):
lgtm

PR #14, opened 2024-01-16
MAP-580 Add capacity and certificate details

Mjwillis:


src/main/resources/db/migration/V1_3__locations.sql
```diff
@@ -0,0 +1,62 @@
+CREATE TABLE capacity
+(
+    id                   uuid not null
```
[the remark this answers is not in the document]
steverendell: And part of the reason I ask is that its a slight shame to have to have a nullable PK in the kotlin

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationDto.kt
```diff
@@ -34,6 +39,42 @@ data class Location(
...
+  @Schema(description = "Alternative description to display for location", example = "Wing A", required = false)
+  val description: String? = null,
+
+  @Schema(description = "Additional comments that can be made about this location", example = "Not to be used", required = false)
+  val comments: String? = null,
+
+  @Schema(description = "Sequence of locations within the current parent location", example = "1", required = false)
+  val orderWithinParentLocation: Int? = null,
+
+  @Schema(description = "Indicates the location is available", example = "true", required = true)
```
danbenton-mojdt, line 51: _Enabled_ rather than _available_, if we want to use a word other that active in the description. IMO a cell could be active but be unavailable for other reasons?

src/main/resources/db/migration/V1_3__locations.sql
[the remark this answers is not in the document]
Mjwillis: No, but I've gone for UUID - but for internal ID's we could use longs

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationDto.kt
Mjwillis: Ok Thanks

Mjwillis: @danbenton-mojdt updated

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Capacity.kt
```diff
@@ -0,0 +1,17 @@
...
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+
+@Entity
+class Capacity(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long? = null,
```
steverendell, line 12: Minor but you could make the ids not-nullable and initialise  to 0. Thats what weve done in other repos. Hibernate understands this value as 'unset'.

steverendell (approved):
lgtm

PR #27, opened 2024-02-14
MAP-636 add apis to get prison internal locations

GurnankCheema:
add apis to access locations within prisons


Mjwillis (approved):
LGTM 
Only comment is that the tests will fix as we change this enums - maybe just check for a few key items.  But can be done if it gets annoying

PR #28, opened 2024-02-16
MAP-791 Upgrade Spring Boot

Mjwillis:
MAP-791 Update library versions in build.gradle.kts

The versions of the Spring Boot plugin, the HMPPS SQS Spring Boot Starter, and the Json Web Token libraries are updated to the latest versions. This is done to stay up-to-date with library patches and updates that may improve performance or fix bugs.


build.gradle.kts
```diff
@@ -22,7 +22,7 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
   implementation("org.springframework.boot:spring-boot-starter-validation")
 
-  implementation("uk.gov.justice.service.hmpps:hmpps-sqs-spring-boot-starter:2.2.1")
+  implementation("uk.gov.justice.service.hmpps:hmpps-sqs-spring-boot-starter:3.1.1")
   implementation("io.opentelemetry:opentelemetry-api:1.34.1")
```
petergphillips, line 26: odd to be at 1.34.1 when app insights is still at 1.32

petergphillips (approved):
LGTM

PR #32, opened 2024-02-22
MAP-801 Refactor attributes representation in Location objects

Mjwillis:
This commit includes changes to refactor how attributes are represented in 'Location' objects across various files. The attributes are now represented as a 'Set' of 'ResidentialAttributeValue' objects rather than a 'Map'. This simplifies the structure of the 'Location' entities.


steverendell (approved):
lgtm

PR #47, opened 2024-03-01
MAP-848 Add functionality to create prison wings

Mjwillis:
This commit introduces a feature for creating prison wings with defined landings and cells. This implementation also includes unit tests for the functions added, ensuring the correct status responses are returned. As part of this new feature, the request body validation has been included to prevent incorrect requests or the creation of duplicate locations. Additionally, a minor typo fix in a variable name from `reactivationDate` to `proposedReactivationDate` was implemented.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
@@ -11,6 +11,7 @@ import org.springframework.context.annotation.Primary
 import org.springframework.security.test.context.support.WithMockUser
 import uk.gov.justice.digital.hmpps.locationsinsideprison.dto.CreateNonResidentialLocationRequest
 import uk.gov.justice.digital.hmpps.locationsinsideprison.dto.CreateResidentialLocationRequest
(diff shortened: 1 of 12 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
```diff
@@ -734,6 +735,115 @@ class LocationResourceIntTest : SqsIntegrationTestBase() {
...
+              "certification": {
+                "certified": true,
+                "capacityOfCertifiedCell": 160
+              }
+            }
+          """,
+            false,
+          )
+      }
+    }
```
steverendell, line 844: I was going to ask if you were intending to read back the created data to verify it, but presumably the endpoint is basically to aid testing rather than being required in prod?

steverendell (approved):
lgtm

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
Mjwillis: I've just read back the capacities to know that all the cells below have been aggregated up.  Doubt we'd use it in prod, but front end might support the "creation" of a new wing

PR #61, opened 2024-03-20
MAP-876 Add parentId to UpsertLocationRequest

Mjwillis:
Enhanced the UpsertLocationRequest class to include a parentId attribute that represents the UUID of the parent location, if it exists. Also updated the related SyncAndMigrateResourceIntTest.kt and SyncService.kt to handle this parameter. This provides an alternative method to associate a location with its parent, offering flexibility in location management processes.


steverendell (approved):
lgtm

PR #64, opened 2024-03-25
MAP-893 Allow deactivated locations to be migrated and cascaded down


Mjwillis:
- Migration need to support deactivation

- deactivation should only be recorded at location specified and will cascade down if parent is inactive

- This affect working capacities and other capacity depending on type of deactivation


steverendell (approved):
lgtm

PR #67, opened 2024-03-26
MAP-896 Add deactivation details and update location indexes


Mjwillis:
Added deactivation details such as deactivatedDate, deactivationReason, and proposedReactivationDate to SyncAndMigrateResourceIntTest and UpsertLocationRequest files. Also created new indexes and dropped the old index in the V1_13__location_indexes.sql file.

In addition to updating the deactivated attribute from active to false for a location, deactivation details are now added to ensure better tracking and management of deactivated locations. Modifying the database indexes is intended to improve query performance related to location and location history.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/UpdateLocationRequest.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/UpdateLocationRequest.kt
@@ -20,4 +20,5 @@ interface UpdateLocationRequest {
   val deactivationReason: DeactivatedReason?
   val proposedReactivationDate: LocalDate?
   val deactivatedDate: LocalDate?
+  fun isDeactivated() = deactivationReason != null && deactivatedDate != null
 }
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/UpdateLocationRequest.kt
```diff
@@ -20,4 +20,5 @@ interface UpdateLocationRequest {
   val deactivationReason: DeactivatedReason?
   val proposedReactivationDate: LocalDate?
   val deactivatedDate: LocalDate?
+  fun isDeactivated() = deactivationReason != null && deactivatedDate != null
```
steverendell, line 23: hmm yes this was worrying me a bit, Ive just checked and there are a lot of rows in Nomis where the 3 deactivated/ reactivate columns are null but active is false too. Might need an explicit 'active' field?

steverendell: Or .. I could set the reason as 'OTHER' in these cases?

Mjwillis: Ah ok, or if not active have a deactivation of OTHER with a last modified date for deactivation?

steverendell (approved):
lgtm

PR #68, opened 2024-03-27
MAP-906 - Show capacity and other attributes at the cell level  even if deactivated. The values DO NOT count towards aggregated values if inactive perm

Mjwillis:


steverendell (approved):
lgtm

PR #70, opened 2024-04-09
MAP-933 Combine Location History with location migration, this only applies to initial migration phase

Mjwillis:


helm_deploy/values-dev.yaml
```diff
@@ -11,6 +11,11 @@ generic-service:
...
     API_BASE_URL_OAUTH: https://sign-in-dev.hmpps.service.justice.gov.uk/auth
 
+  scheduledDowntime:
+    enabled: true
+    startup: '49 6 * * 1-5' # Start at 6.49am UTC Monday-Friday
+    shutdown: '58 21 * * 1-5' # Stop at 9.58pm UTC Monday-Friday
+
 # CloudPlatform AlertManager receiver to route prometheus alerts to slack
 # See https://user-guide.cloud-platform.service.justice.gov.uk/documentation/monitoring-an-app/how-to-create-alarms.html#creating-your-own-custom-alerts
 generic-prometheus-alerts:
```
steverendell, line 21: There is a 'businessHoursOnly: true' setting just below here which you might need in preprod too to inhibit any alerts resulting from stuff being down.

steverendell (approved):

PR #75, opened 2024-04-22
MAP-971 Add ability to delete a location (sync only from NOMIS)

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/SyncService.kt
```diff
@@ -12,6 +12,7 @@ import uk.gov.justice.digital.hmpps.locationsinsideprison.jpa.Location
 import uk.gov.justice.digital.hmpps.locationsinsideprison.jpa.repository.LocationRepository
 import uk.gov.justice.digital.hmpps.locationsinsideprison.resource.LocationNotFoundException
 import java.time.Clock
+import java.util.*
```
steverendell, line 15: I get this as well in my intellij - it wont seem to just add java.util.UUID !

steverendell (approved):
lgtm

PR #91, opened 2024-05-08
MAP-871 Add prisoner search service and prevent location deactivation with prisoners

Mjwillis:
This update includes the implementation of prisoner search service, as well as the prevention of location deactivation if prisoners are present. A check has been added before deactivating a location to ensure it does not contain any prisoners. If found, the deactivation process is stopped and an error message is returned.


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/integration/wiremock/PrisonerSearchMockServer.kt
```diff
@@ -0,0 +1,87 @@
...
+    )
+  }
+
+  fun stubSearchByLocations(
+    prisonId: String,
+    locations: List<String>,
+    returnResult: Boolean = false,
+  ) {
+    val requestBody = mapper.writeValueAsString(
+      /* value = */
```
steverendell, line 43: ^ probably not meant to be left in?

steverendell (approved):
lgtm

PR #92, opened 2024-05-09
MAP-1091 Removed the certification exception for checking cell capacity, as there is no longer a need to have this unnecessary check.

Mjwillis:


steverendell (approved):
lgtm

PR #104, opened 2024-05-15
MAP-1095 add endpoint to get locations by type

GurnankCheema:


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
```diff
@@ -162,6 +162,164 @@ class LocationResourceIntTest : SqsIntegrationTestBase() {
     repository.save(wingB)
   }
 
+  @DisplayName("GET /locations")
```
GurnankCheema, line 165: moved this test from lower down to first place as it's the first endpoint

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
```diff
@@ -1050,123 +1207,30 @@ class LocationResourceIntTest : SqsIntegrationTestBase() {
...
-                  "pageNumber": 0,
-                  "paged": true,
-                  "unpaged": false
-                },
-                "empty": false
-              }
-
-              """,
+             [{
+              "prisonId": "MDI",
```
GurnankCheema, line 1219: only included the attributes we are interested in

Mjwillis (approved):
LGTM

PR #105, opened 2024-05-16
MAP-1095 new endpoint to get active locations by their path hierarchy

GurnankCheema:
ticket:
_POST prisonLocationsByKeys(keys[]): Location[], to get the location descriptions for a list of keys provided in one call, for when we show lists of video bookings. Useful if this could avoid being specific to one prison because there could be a use-case where one video booking could have 2 or more prisoners, in different prisons, as co-defendants._ 

btw, should locationResourceIntTest.kt Line 160 say landing1 as the inactive cell is on landing 1 not landing 3?

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
@@ -169,8 +169,6 @@ class LocationResource(
     @RequestBody
     @Validated
     keys: List<String>,
-    @RequestParam(name = "includeChildren", required = false, defaultValue = "false") includeChildren: Boolean = false,
-    @RequestParam(name = "includeHistory", required = false, defaultValue = "false") includeHistory: Boolean = false,
   ): List<LocationDTO> = locationService.getLocationsByKeys(keys)
 
   @GetMapping("/prison/{prisonId}")
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
@@ -132,6 +132,47 @@ class LocationResource(
   ) =
     locationService.getLocationByKey(key = key, includeChildren = includeChildren, includeHistory = includeHistory) ?: throw LocationNotFoundException(key)
 
+  @PostMapping("/keys", produces = [MediaType.APPLICATION_JSON_VALUE])
+  @PreAuthorize("hasRole('ROLE_VIEW_LOCATIONS')")
+  @ResponseStatus(HttpStatus.OK)
+  @Operation(
+    summary = "Gets locations by their keys",
+    description = "Requires role VIEW_LOCATIONS",
+    responses = [
+      ApiResponse(
+        responseCode = "200",
+        description = "Returns location",
+      ),
+      ApiResponse(
+        responseCode = "400",
+        description = "Invalid Request",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "401",
+        description = "Unauthorized to access this endpoint",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "403",
+        description = "Missing required role. Requires the MAINTAIN_LOCATIONS role with write scope.",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "404",
+        description = "Data not found",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  fun getLocationsByKeys(
+    @RequestBody
+    @Validated
+    keys: List<String>,
+    @RequestParam(name = "includeChildren", required = false, defaultValue = "false") includeChildren: Boolean = false,
+    @RequestParam(name = "includeHistory", required = false, defaultValue = "false") includeHistory: Boolean = false,
+  ): List<LocationDTO> = locationService.getLocationsByKeys(keys, includeChildren, includeHistory)
+
   @GetMapping("/prison/{prisonId}")
   @ResponseStatus(HttpStatus.OK)
   @PreAuthorize("hasRole('ROLE_VIEW_LOCATIONS')")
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
```diff
@@ -940,6 +940,109 @@ class LocationResourceIntTest : SqsIntegrationTestBase() {
...
+
+    @Nested
+    inner class HappyPath {
+
+      @Test
+      fun `can retrieve locations from a list of keys`() {
+        webTestClient.post().uri("/locations/keys?includeChildren=true")
+          .headers(setAuthorisation(roles = listOf("ROLE_VIEW_LOCATIONS")))
+          .header("Content-Type", "application/json")
+          .bodyValue(listOf("B-A", "B-A-001", "MDI-Z", "Z", "Z-2", "Z-VISIT", "Z-1-003", "XYZ-1-2-3"))
```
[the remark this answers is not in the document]
GurnankCheema: you are right -  it doesn't and is not included in the .expectBody().json(...

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
```diff
@@ -940,6 +940,120 @@ class LocationResourceIntTest : SqsIntegrationTestBase() {
...
```
Mjwillis, line 988: So if there is no match then it just ignores the request?

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
```diff
@@ -132,6 +132,47 @@ class LocationResource(
...
+        description = "Data not found",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  fun getLocationsByKeys(
+    @RequestBody
+    @Validated
+    keys: List<String>,
+    @RequestParam(name = "includeChildren", required = false, defaultValue = "false") includeChildren: Boolean = false,
```
Mjwillis, line 172: Don't think we need to return child locations or history

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -77,6 +77,49 @@ class LocationService(
...
+    }
+
+    val queryResponse = mutableListOf<Location>()
+    locationsGroupedByPrisonId.forEach { it ->
+      queryResponse += locationRepository.findAllByPrisonIdAndPathHierarchyIsIn(it.prisonId, it.pathHierarchy)
+    }
+
+    return queryResponse
+      .filter { it.isActive() }
+      .map { it.toDto(includeChildren = includeChildren, includeHistory = includeHistory) }
```
Mjwillis, line 119: I would include children or history - no requirement for this and also there are other endpoint for this information

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -77,6 +77,49 @@ class LocationService(
...
+      locationsGroupedByPrisonId += PrisonAndPathHierarchies(prisonId, pathHierarchies)
+    }
+
+    val queryResponse = mutableListOf<Location>()
+    locationsGroupedByPrisonId.forEach { it ->
+      queryResponse += locationRepository.findAllByPrisonIdAndPathHierarchyIsIn(it.prisonId, it.pathHierarchy)
+    }
+
+    return queryResponse
+      .filter { it.isActive() }
```
Mjwillis, line 118: I wouldn't filter actives either

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -77,6 +77,49 @@ class LocationService(
...
+        if (location.prisonId == prisonId) {
+          pathHierarchies += location.pathHierarchy
+        }
+      }
+      locationsGroupedByPrisonId += PrisonAndPathHierarchies(prisonId, pathHierarchies)
+    }
+
+    val queryResponse = mutableListOf<Location>()
+    locationsGroupedByPrisonId.forEach { it ->
+      queryResponse += locationRepository.findAllByPrisonIdAndPathHierarchyIsIn(it.prisonId, it.pathHierarchy)
```
Mjwillis, line 114: This could be a bit slow as it will make individual call.  I think there are two options:
1. Just support locations from one prison at a time, (I can't think why they'd need to do it across prisons)
2. Write some JQL instead and do it one query e.g. concat two fields

Mjwillis: e.g. 
```sql
select l.* from location l
where concat(prison_id,'-',path_hierarchy) IN ('LEI-MW', 'MDI-1')
```

Mjwillis: @GurnankCheema 
Think this should do it in repo 
```kotlin
  @Query("select l from Location l where concat(l.prisonId,'-',l.pathHierarchy) IN (:keys)")
  fun findAllByKeys(keys: List<String>): List<Location>
```

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
GurnankCheema: yes

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
GurnankCheema: wasn't sure but will remove

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
GurnankCheema: I will use the JQL as it gives us more flexibility for the future

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
Mjwillis, line 172: Don't need these

Mjwillis (changes requested):
One more change in the API contract

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
GurnankCheema: doh!

Mjwillis (approved):
LGTM

PR #112, opened 2024-05-24
MAP-1142 Add tests for update and reactivation for location properties

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationDto.kt
```diff
@@ -111,6 +111,9 @@ data class Location(
   @Schema(description = "Proposed Date for location reactivation", example = "2026-01-24", required = false)
   val proposedReactivationDate: LocalDate? = null,
 
+  @Schema(description = "Planet FM Reference", example = "2323/45M", required = false)
```
Mjwillis, line 114: This was missing from DTO!

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -364,7 +364,7 @@ class LocationService(
       null,
     )
 
-    return locationToUpdate.toDto(includeParent = true)
+    return locationToUpdate.toDto(includeParent = true, includeChildren = reactivateSubLocations)
```
Mjwillis, line 367: If you cascade the re-activation you need to send events for reactivation of sub locations

Thource (approved):

PR #124, opened 2024-05-31
MAP-1169 Performance improvements

Mjwillis:


Thource (approved):

src/main/resources/db/migration/V1_26__cascade_delete.sql
```diff
@@ -0,0 +1,48 @@
+alter table location
```
Mjwillis, line 1: I've added the cascade to delete a single object but will only delete if no children below.

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/SyncAndMigrateResourceIntTest.kt
```diff
@@ -895,6 +895,14 @@ class SyncAndMigrateResourceIntTest : SqsIntegrationTestBase() {
           .exchange()
           .expectStatus().is4xxClientError
       }
+
+      @Test
+      fun `cannot delete an existing location which as children`() {
+        webTestClient.delete().uri("/sync/delete/${wingB.id}")
```
Mjwillis, line 907: This test checks that locations with children cannot be deleted

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Location.kt
```diff
@@ -70,9 +71,11 @@ abstract class Location(
   open var proposedReactivationDate: LocalDate? = null,
   open var planetFmReference: String? = null,
 
-  @OneToMany(mappedBy = "parent", fetch = FetchType.LAZY, cascade = [CascadeType.ALL])
+  @BatchSize(size = 1000)
+  @OneToMany(mappedBy = "parent", fetch = FetchType.EAGER, cascade = [CascadeType.ALL])
```
Mjwillis, line 75: This is the biggy,  it turns 1000's of SQL calls into about 5

PR #125, opened 2024-05-31
MAP-1172 UI improvements

Mjwillis:
- Return level of locations
- Return the heirarchy of a location
- Return a history of changes (last 10) for the current location

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationDto.kt
@@ -109,6 +109,12 @@ data class Location(
   @Schema(description = "Top Level Location Id", example = "57718979-573c-433a-9e51-2d83f887c11c", required = true)
   val topLevelId: UUID,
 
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationDto.kt
```diff
@@ -109,6 +109,12 @@ data class Location(
   @Schema(description = "Top Level Location Id", example = "57718979-573c-433a-9e51-2d83f887c11c", required = true)
   val topLevelId: UUID,
 
+  @Schema(description = "Current Level with hierarchy", example = "1", required = true)
```
danbenton-mojdt, line 112: Worth noting it's zero-indexed?

Mjwillis: It starts at 1. Wing = 1, Landing =2, Cell=3

Mjwillis: But yes could explain more

Thource (approved):

PR #130, opened 2024-06-03
MAP-1178 Inactive count incorrectly calculated.

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -635,7 +635,7 @@ class LocationService(
     return (
       startLocation?.findAllLeafLocations() ?: cellLocationRepository.findAllByPrisonIdAndActive(prisonId, false)
       )
-      .filter { !it.isPermanentlyDeactivated() && !it.isActiveAndAllParentsActive() }
```
marcusaleman, line 638: open fun isTemporarilyDeactivated() = !isActiveAndAllParentsActive() && !isPermanentlyDeactivated()

Mjwillis: Yes it calls isTemporarilyDeactivated

PR #134, opened 2024-06-05
MAP-1200 ✨ Added custom UUID generation for Location entity

Mjwillis:
A new custom UUID generator, UuidV7Generator, has been added to the project. The change replaces the previous random UUID generation strategy in the Location entity with this custom time-based UUID v7 generation. A new dependency for the java-uuid-generator library has been added to the project build configuration.


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/helper/UuidV7Generator.kt
```diff
@@ -0,0 +1,38 @@
+package uk.gov.justice.digital.hmpps.locationsinsideprison.jpa.helper
```
Mjwillis, line 1: Copied off incident reporting code

ushkarev (approved):

PR #143, opened 2024-06-12
MAP-1030 changed dto to use code and do sort

GurnankCheema:
refactored so that dto has same data and structure as the corresponding dto in whereaboutsApi
The two should match but currently in the case of the location api, the dto didn't sort children and the key value for the group didn't match


michalnawrockidj (approved):
LGTM

marcusaleman (commented):
+

PR #149, opened 2024-06-19
MAP-1030 provide alternative value for dto.name

GurnankCheema:
This is to display an alternative when 'localName'  has not been set for a wing/landing/spur


GurnankCheema:
thanks @Mjwillis

PR #155, opened 2024-06-20
Add location type of BLOCK

danbenton-mojdt:


Mjwillis (commented):
BLOCK isn't a thing, there is a houseblock but we again not use it as a residential type.  Lets chat about what the problem is

PR #167, opened 2024-07-03
MAP-1371 Provide cells with capacity endpoint 

Mjwillis:
This replaces the prison API and whereabouts API endpoints that find capacity of cells within prison.

It makes use of prisoner search to look up locations and check if prisoners are in the cells.  Capacity is then used to work out if space is available based on existing logic.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationOccupancyResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationOccupancyResource.kt
@@ -1,0 +1,121 @@
+package uk.gov.justice.digital.hmpps.locationsinsideprison.resource
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.media.Content
+import io.swagger.v3.oas.annotations.media.Schema
+import io.swagger.v3.oas.annotations.responses.ApiResponse
+import io.swagger.v3.oas.annotations.tags.Tag
+import jakarta.validation.constraints.Pattern
+import jakarta.validation.constraints.Size
+import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
+import org.springframework.security.access.prepost.PreAuthorize
+import org.springframework.validation.annotation.Validated
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.ResponseStatus
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.locationsinsideprison.jpa.SpecialistCellType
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationOccupancyResource.kt
```diff
@@ -0,0 +1,121 @@
...
+) {
+  @Schema(description = "Business Key for a location", example = "MDI-A-1-001", required = true)
+  fun getKey(): String {
+    return "$prisonId-$pathHierarchy"
+  }
+
+  @JsonIgnore
+  fun hasSpace() = noOfOccupants < getActualCapacity()
+
+  private fun getActualCapacity() = if (workingCapacity != 0) workingCapacity else maxCapacity
```
Mjwillis, line 120: This is the same logic as prison-api

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -672,6 +684,53 @@ class LocationService(
...
+    } else {
+      cellLocationRepository.findAllByPrisonIdAndActive(prisonId, true)
+    }
+
+    val cellsWithCapacity = cellsInGroup(
+      prisonId = prisonId,
+      groupName = groupName,
+      cellsToFilter = cellsToFilter,
+    )
+      .filter { cell -> cell.isActiveAndAllParentsActive() && cell.isCertified() }
```
Mjwillis, line 707: Additional "certified" check not present in Prison API which NOMIS does apply

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationOccupancyResource.kt
```diff
@@ -0,0 +1,121 @@
...
+    @PathVariable prisonId: String,
+    @Schema(description = "Location Id in the prison below which to find cells", example = "de91dfa7-821f-4552-a427-bf2f32eafeb0", required = false)
+    @RequestParam(name = "locationId", required = false) locationId: UUID? = null,
+    @Schema(description = "Group name for a sub location to find cells", example = "Wing A", required = false)
+    @RequestParam(name = "groupName", required = false) groupName: String? = null,
+    @Schema(description = "Only return cells of a specified specialist type", example = "CSU", required = false)
+    @RequestParam(name = "specialistCellType", required = false) specialistCellType: SpecialistCellType? = null,
+    @Schema(description = "Include prisoner details in this cell", required = false, defaultValue = "false")
+    @RequestParam(name = "includePrisonerInformation", required = false, defaultValue = "false") includePrisonerInformation: Boolean = false,
+  ): List<CellWithSpecialistCellTypes> =
```
marcusaleman, line 80: is it not required to publish an event for this?

marcusaleman (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationOccupancyResource.kt
Mjwillis: No because it doesn't change any data.  Only "writes" cause an event

PR #171, opened 2024-07-08
[MAP-1359] Add Kover plugin to generate code coverage reports

jnioche-jd:
Introduce a test coverage plugin to allow developers to track their testing

Documentation:
https://kotlin.github.io/kotlinx-kover/gradle-plugin/

Example command to generate a HTML report locally:
`./gradlew koverHtmlReport`

Example report output:

image

Later on we can look at ways of generating these automatically with CircleCI and adopt it in our other projects using Kotlin.


marcusaleman (approved):

Mjwillis (approved):
Would you wire that into the circle build too?

jnioche-jd:
> Would you wire that into the circle build too?

I think so. Could be done as a subtask of MAP-1395

PR #178, opened 2024-07-10
Map 1350 locations api create new prefix endpoint

GurnankCheema:
this will replace the call clients currently make to whereaboutApi's GET /locations/{agencyId}/{group}/location-prefix


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/ApiExceptionHandler.kt
```diff
@@ -313,6 +326,9 @@ class ApiExceptionHandler {
 }
 
 class LocationNotFoundException(id: String) : Exception("There is no location found for ID = $id")
+
+class LocationPrefixNotFoundException(id: String) : Exception("No mappings found for $id")
```
Mjwillis, line 330: Think message could be clearer

Mjwillis (approved):
One minor comment.

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/ApiExceptionHandler.kt
GurnankCheema: will change to "Location prefix not found for $id"

Mjwillis (approved):
LGTM

marcusaleman (commented):
seen.

PR #181, opened 2024-07-10
MAP-1395 Separate endpoint for updating specialist cell types to allow removal of all types

jnioche-jd:
Initial version, still need to add to the validation logic, see [issue](https://dsdmoj.atlassian.net/browse/MAP-1395?focusedCommentId=480549)

Adds a new endpoint and removes the update of the specialist cell types as part of **PATCH  /locations/residential/{id}**

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
@@ -366,6 +366,40 @@ class LocationService(
   }
 
   @Transactional
+  fun updateSpecialistCellTypes(id: UUID, specialistCellTypes: Set<SpecialistCellType>): LocationDTO {
+    val cell = cellLocationRepository.findById(id)
+      .orElseThrow { LocationNotFoundException(id.toString()) }
+
+    if (cell.isPermanentlyDeactivated()) {
+      throw ValidationException("Cannot change the specialist cell types of a permanently deactivated location")
+    }
+
+    // TODO check for empty sets when 0 working cap etc (e.g. validation rules)
+//    val prisoners = prisonerLocationService.prisonersInLocations(cell)
+//    if (maxCapacity < prisoners.size) {
+//      throw CapacityException(locCapChange.getKey(), "Max capacity ($maxCapacity) cannot be decreased below current cell occupancy (${prisoners.size})")
+//    }
+
+    cell.updateSpecialistCellTypes(
+      specialistCellTypes,
+      authenticationFacade.getUserOrSystemInContext(),
+      clock,
+    )
+    log.info("Updated specialist cell types = $specialistCellTypes")
+
+    telemetryClient.trackEvent(
+      "Specialist cell types updated",
+      mapOf(
+        "id" to id.toString(),
+        "key" to cell.getKey(),
+        "specialistCellTypes" to specialistCellTypes.toString(),
+      ),
+      null,
+    )
+    return cell.toDto(includeParent = false, includeNonResidential = false)
+  }
+
+  @Transactional
   fun updateLocation(id: UUID, updateLocationRequest: UpdateLocationRequest): LocationDTO {
     val location = locationRepository.findById(id)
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -383,6 +383,40 @@ class LocationService(
...
+  @Transactional
+  fun updateSpecialistCellTypes(id: UUID, specialistCellTypes: Set<SpecialistCellType>): LocationDTO {
+    val cell = cellLocationRepository.findById(id)
+      .orElseThrow { LocationNotFoundException(id.toString()) }
+
+    if (cell.isPermanentlyDeactivated()) {
+      throw ValidationException("Cannot change the specialist cell types of a permanently deactivated location")
+    }
+
+    // TODO check for empty sets when 0 working cap etc (e.g. validation rules)
```
Mjwillis, line 395: I assume you are coming back to this.  The validation can but pushed into the domain model too

Mjwillis: The rule is already there in code, 

https://github.com/ministryofjustice/hmpps-locations-inside-prison-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Cell.kt#L236-L238

Maybe pull this out to be called

Mjwillis (approved):
Looks good, 

I'd add the validation in next PR and also test this scenario, e.g. removing speciallist cell types

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
jnioche-jd: Thanks. I have pushed a new commit that should implement the logic.

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
```diff
@@ -1987,6 +1987,153 @@ class LocationResourceIntTest : SqsIntegrationTestBase() {
...
+        val result = webTestClient.put().uri("/locations/${cell1.id}/specialist-cell-types")
+          .headers(setAuthorisation(roles = listOf("ROLE_MAINTAIN_LOCATIONS"), scopes = listOf("write")))
+          .header("Content-Type", "application/json")
+          .bodyValue(jsonString(setOf(SpecialistCellType.BIOHAZARD_DIRTY_PROTEST)))
+          .exchange()
+          .expectStatus().isOk
+          .expectBody(LocationTest::class.java)
+          .returnResult().responseBody!!
+
+        assertThat(result.usedFor == expectedUsedFor)
```
danbenton-mojdt, line 2104: I think you want to check specialist cell type rather than used for? The variable name above could change too

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResourceIntTest.kt
```diff
@@ -1987,6 +1987,153 @@ class LocationResourceIntTest : SqsIntegrationTestBase() {
...
+        val result = webTestClient.put().uri("/locations/${cell2.id}/specialist-cell-types")
+          .headers(setAuthorisation(roles = listOf("ROLE_MAINTAIN_LOCATIONS"), scopes = listOf("write")))
+          .header("Content-Type", "application/json")
+          .bodyValue(jsonString(emptySet<SpecialistCellType>()))
+          .exchange()
+          .expectStatus().isOk
+          .expectBody(LocationTest::class.java)
+          .returnResult().responseBody!!
+
+        assertThat(result.usedFor == expectedUsedFor)
```
danbenton-mojdt, line 2126: Ditto specialist cell type rather than usedFor

danbenton-mojdt (changes requested):

PR #186, opened 2024-07-18
MAP-1418 Refactor exception handling and introduce error codes

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -404,12 +407,12 @@ class LocationService(
...
 
     if (cell.isPermanentlyDeactivated()) {
-      throw ValidationException("Cannot change the specialist cell types of a permanently deactivated location")
+      throw PermanentlyDeactivatedUpdateNotAllowedException(cell.getKey())
     }
 
     // Check that the workingCapacity is not set to 0 for normal accommodations when removing the specialists cell types
-    if (specialistCellTypes.isEmpty() && cell.accommodationType.equals(AccommodationType.NORMAL_ACCOMMODATION) && cell.getWorkingCapacity() == 0) {
-      throw ValidationException("Cannot removes specialist cell types for a normal accommodation with a working capacity of 0")
+    if (specialistCellTypes.isEmpty() && cell.accommodationType == AccommodationType.NORMAL_ACCOMMODATION && cell.getWorkingCapacity() == 0) {
```
Mjwillis, line 414: @jnioche-jd in kotlin you don't need to use `.equals` for strings (or enums)

jnioche-jd: As a Java dev, I find this deeply offensive :-)

PR #192, opened 2024-07-24
Map 1402

marcusaleman:
Create common data to be used in the splitted tests. 
Create separate test file based on the type of HTTP request method (PUT).


marcusaleman:
@jnioche-jd  @danbenton-mojdt

**Plan** :  (if you have any other approachment let me know this afternoon) -> i couldn't connect this morning .
1) Need to fix two test.
2) Split request for GET and POST files as I did for PUT
3) PR final for reviewing and merge main (all validations passed)

4) Create a ticket to spike this which should be straight forward. This inject the common data used for the integration tests.
@Sql("classpath:repository/vacated-prisoner-multiple-residences.sql") 
as @michalnawrockidj has suggested.

PR #193, opened 2024-07-29
MAP-1402 split large test class into smaller files

marcusaleman:
LocationResourceintTest.kt has been splitted in functional smaller pieces.

I found using a supertype a way to not duplicate code as mostly of the Data used in the test is common.


jnioche-jd:
@marcusaleman the main branch runs 261 tests, this one 276. Is this expected? 

Could we improve the package and file names? At the moment we get paths like 

`locationsinsideprison/resource/locationresource/LocationResourceKeyTest.kt`

with loads of redundancy in the names.

What about  for instance `.../resource/location/KeyTest.kt` and similar for the other files?

or alternatively, removing them from the package, keep the file names as they are and have them alongside the other tests in `locationsinsideprison/resource/`?

marcusaleman:
@jnioche-jd 
1.) I will refactor the name redundancy
2) I want to believe that tests number should be same. i will check if i left any duplicated. good catch.
3) Alternatively i am spending some time in a  TEST DATA strategy injecting SQL . let's see if i get positive result for a POC (Proof of concept) in 1 integration test at least (subset)

marcusaleman:
@jnioche-jd 
1) please review . I have addressed your suggestions/comments. I fixed the 1 test gap . thx

jnioche-jd:
> @jnioche-jd
> 
> 1. please review . I have addressed your suggestions/comments. I fixed the 1 test gap . thx

There is still redundancy in the path names: the package names ends in /location and the files names start with Location. Maybe just move the files up to the parent dir (resource/) keeping the names as they are. Also this way you shouldn't need the change in the imports for EXPECTED_USERNAME.

When calling `./gradlew clean test` I am getting

` SUCCESS: Executed 277 tests in 1m 21s`

on this branch vs

`SUCCESS: Executed 261 tests in 1m 19s
`
on main.

Any clues?

marcusaleman:
@jnioche-jd 
I get 277 test in **main** and **Map-1402**

jnioche-jd:
a fresh fetch and pull and I am now getting 277 in both branches. all good on that front

Mjwillis (approved):
LGTM

PR #194, opened 2024-07-29
MAP-1402 test data not modified

marcusaleman:
They are the same Tests but I have maintain the Data every file 

Files
LocationResourceIntTest.kt
LocationResourceKeyTest.kt
LocationResourcePrisonIdTest
LocationResourceTransformTest.kt
LocationResourceResidentialTest.kt


jnioche-jd:
#193 feels like a better approach. There is nothing to gain from repeating the same code across every single test class

marcusaleman:
@jnioche-jd @danbenton-mojdt 
right, I agree but i was unsure we want using supertype/inheritance . 

**secondlyI** am going to do a POC injecting SQL in the test and then i will connect with you in a call in order to deliver this ticket.

PR #195, opened 2024-07-30
MAP-1438 locations convert to cell test scenarios and Logic fix for Healthcare inpatients and Care and Speration

marcusaleman:
When receiving a request to convert a location to a cell and the AccommodationType is any of, capacity has to be updated accordingly to a RESIDENTIAL CELL.

Normal accommodation
Healthcare inpatients
Care and separation

WheN AccommodationType = OthersNonResidential capacity is set 0. this has to be discussed I have created ticket MAP-1445.


marcusaleman:
SUCCESS: Executed 280 tests in 1m 18s

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Cell.kt
```diff
@@ -194,7 +194,10 @@ class Cell(
       addSpecialistCellType(it, userOrSystemInContext, clock)
     }
 
-    if (accommodationType == AccommodationType.NORMAL_ACCOMMODATION) {
+    if (accommodationType == AccommodationType.NORMAL_ACCOMMODATION ||
```
danbenton-mojdt, line 197: I might have checked that it's not OTHER rather than checking each of the other options, but it achieves the same thing. This is perhaps just a little more computationally expensive.

danbenton-mojdt:
Minor comment about efficiency but other than that looks good, nice one

danbenton-mojdt (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Cell.kt
jnioche-jd: ... and harder to read

marcusaleman: @danbenton-mojdt right but I don't know exactly what Business want to do with OTHER... 
A) is a Non Res and it can be a CELL? and capacity is 0. but it might have? 
B) it should not be a CELL ?
I have created a ticket for OTHER if at least we can discuss that with the Business . by default i would leave in this way.

Mjwillis: can just use `in`

Mjwillis: I added OTHER as placeholder for cells I didn't know what to do with as part of migration.  We need to not allow OTHER and definitely now allow this API call to set it in this context.  I think the Enum will have to change to only allow the 3 other types.

Mjwillis (changes requested):
prevent OTHER being passed to the API

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Cell.kt
marcusaleman: @Mjwillis  We can do 'Enum will have to change to only allow the 3 other types.'under MAP-1445' for now I commit what agreed in MAP-1438 (this PR)

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Cell.kt
```diff
@@ -194,7 +194,8 @@ class Cell(
       addSpecialistCellType(it, userOrSystemInContext, clock)
     }
 
-    if (accommodationType == AccommodationType.NORMAL_ACCOMMODATION) {
```
Mjwillis, line 197: Are you changing the allowed enum types in this PR or another one?

Mjwillis (changes requested):
Change the request to only show the 3 types not other

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Cell.kt
marcusaleman: hi @Mjwillis just. I just changed the the condition. A) I changed the allowed enum using 'in' in this PR this piece of code is old **however** I haven't work the OTHER in the API in this code. (just I have seen the comment from you right now.

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Cell.kt
marcusaleman: @Mjwillis  I have API call not to pass OTHER however i can adress Enum later which is simpler.. ..

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Cell.kt
marcusaleman: @Mjwillis @danbenton-mojdt 
+ I have prevent API to accept OTHER.
+ I have not modified ENUM AccommodationType due some transaction/changes in other places else but maybe it is the right way to .

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
```diff
@@ -1167,6 +1172,9 @@ class LocationResource(
     return eventPublishAndAudit(
       InternalLocationDomainEventType.LOCATION_AMENDED,
     ) {
+      if (convertToCellRequest.accommodationType == AccommodationType.OTHER_NON_RESIDENTIAL) {
```
Mjwillis, line 1175: Would have prefer to not allow in validation section but its ok

Mjwillis (approved):

PR #211, opened 2024-08-07
MAP-1457 52 Syncservice getLegacyLocation true false coverage 

marcusaleman:
-> Note I am creating a Resource test to cover this Service scenario.
Class | method | line | Branch
**SyncService.kt**
0: 50 76 93 83
1: 50 76 93 **85** **this commit**
SUCCESS: Executed 282 tests in 1m 13s (+2 tests)
+ Draft status
+ It covers SyncService.getLegacyLocation().
+ I might refactor these tests to other class

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/SyncAndMigrateResourceIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/SyncAndMigrateResourceIntTest.kt
@@ -1,15 +1,19 @@
 package uk.gov.justice.digital.hmpps.locationsinsideprison.resource
 
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Assertions
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.DisplayName
 import org.junit.jupiter.api.Nested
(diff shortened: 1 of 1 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/SyncAndMigrateResourceIntTest.kt
```diff
@@ -132,6 +137,47 @@ class SyncAndMigrateResourceIntTest : SqsIntegrationTestBase() {
     repository.save(wingB)
   }
 
+  @Nested
```
Mjwillis, line 140: This is an integration test not a service test, please move to an new test class

Mjwillis (changes requested):
Move to another test class

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/SyncAndMigrateResourceIntTest.kt
Mjwillis: Also you could write an integration test (calling API) to cover these off too

Mjwillis (changes requested):
Add tests at integration level too

PR #219, opened 2024-08-13
MAP-1496 Sort location attributes by description


Mjwillis:


marcusaleman (approved):
LGTM

jimbali:
@Mjwillis this means that where we have a bunch of options and also "other" (which might have a free text field), "other" will not appear at the bottom of the list. I think the designers are expecting a certain ordering of the fields. It might be worth asking them if they want "other" to be at the bottom or not.

PR #220, opened 2024-08-13
MAP-1457-52-Sync-Resource-Check-gaps-in-test-coverage-for-Kotlin-project-location-api

marcusaleman:


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/SyncAndMigrateResourceIntTest.kt
```diff
@@ -1020,4 +1050,95 @@ class SyncAndMigrateResourceIntTest : SqsIntegrationTestBase() {
...
+      fun `cannot sync an existing location`() {
+        webTestClient.get().uri("/sync/id/${wingB.id}")
+          .headers(setAuthorisation(roles = listOf("ROLE_SYNC_LOCATIONS"), scopes = listOf("write")))
+          .exchange()
+          .expectStatus().is4xxClientError
+      }
+    }
+
+    @Nested
+    inner class HappyPath {
```
Mjwillis, line 1112: Not sure why you are mocking out repository and service?  Why not call the service as normal as there is data there for cell1 and it will be returned in Legacy format.  Also there is a test for /sync/id/ already but happy to add more.

Mjwillis (changes requested):
Remove mocks, add full integration flow

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/SyncAndMigrateResourceIntTest.kt
marcusaleman: @Mjwillis 
1. The idea is touching this method SyncService.**getLegacyLocation**.
2. I need to get _legacylocation_object returned from this service SyncSerivce.getlegacylocation    
	-> I want to believe that I should beable to add this to CommonDataTestBase or add it in the setup()
	else
	I would leave this Mock as it PASS.

marcusaleman: @Mjwillis 
I have amended/corrected using SynService directly as it suggested (it is better option)
I have deleted repository as it is not needed.
It has increased Positive Hit for **fun getLegacyLocation**

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/SyncAndMigrateResourceIntTest.kt
```diff
@@ -1020,4 +1027,93 @@ class SyncAndMigrateResourceIntTest : SqsIntegrationTestBase() {
...
+
+      @Test
+      fun `can retrieve sync Nomis to an existing location`() {
+        legacylocation = syncService.getLegacyLocation(id = cell.id!!, includeHistory = false)!!
```
Mjwillis, line 1093: Dont' need to load the legacy location as the cell field has this information

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/SyncAndMigrateResourceIntTest.kt
```diff
@@ -1020,4 +1027,93 @@ class SyncAndMigrateResourceIntTest : SqsIntegrationTestBase() {
...
+      @Test
+      fun `handles location not found`() {
+        webTestClient.get().uri("/sync/id/${UUID.randomUUID()}")
+          .headers(setAuthorisation(roles = listOf("ROLE_SYNC_LOCATIONS"), scopes = listOf("write")))
+          .exchange()
+          .expectStatus().is4xxClientError
+      }
+
+      @Test
+      fun `cannot sync an existing location`() {
```
Mjwillis, line 1080: This test doesn't make sense, what it testing? This is not syncing a location.  Why does it error with 4xx?

Mjwillis (changes requested):
- Remove Line 70
- Rework test as show in review

marcusaleman:
I have amended with comments as they make code more concise. push code changes. thanks

Mjwillis (approved):
LGTM

PR #231, opened 2024-08-21
MAP1457  Resource Location id not found validation test

marcusaleman:
As Integration Tests is not cover /locations/id when id is not found . this is a validation test for that scenario.


Mjwillis (approved):
LGTM

PR #241, opened 2024-08-24
MAP-1457 112 LocationGroupDto class PrisonId Groups Resource to standard Test

marcusaleman:
As part of better Kotlin Test practices.
Get rid of Json strings in the test
Use a Data Class to get the webservice result/response answer


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationGroupDtoListTest.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.locationsinsideprison.dto
+
+import com.fasterxml.jackson.annotation.JsonInclude
+
+@JsonInclude(JsonInclude.Include.ALWAYS)
```
Mjwillis, line 5: This class should be a data class and also included in the `LocationPrisonIdResourceTest.kt` file instead

Mjwillis (changes requested):
Move class and change to data class

Mjwillis (approved):
lgtm

PR #244, opened 2024-08-27
MAP-1564 Add all demo prison locations in SQL script

Mjwillis:


GurnankCheema (approved):
LGTM

marcusaleman (approved):

src/main/resources/db/training/R__03_training_prison_locations.sql
```diff
@@ -1,5 +1,126 @@
+call setup_prison_demo_locations('AGI', 'ZQW42M');
+call setup_prison_demo_locations('ACI', 'ZQW42M');
+call setup_prison_demo_locations('AYI', 'ZQW42M');
+call setup_prison_demo_locations('BFI', 'ZQW42M');
+call setup_prison_demo_locations('ASI', 'ZQW42M');
+call setup_prison_demo_locations('BAI', 'ZQW42M');
+call setup_prison_demo_locations('BWI', 'ZQW42M');
+call setup_prison_demo_locations('BMI', 'ZQW42M');
+call setup_prison_demo_locations('BSI', 'ZQW42M');
+call setup_prison_demo_locations('BLI', 'ZQW42M');
```
marcusaleman, line 10: Would be a good idea to use the Test data from the training environment into the tests (injecting them as SQL , etc) or it doesn't worth the effort at all to to that.. ?

Mjwillis: I've done it for the docker environment.  There is too much data probably for test data and will slow down loading

PR #245, opened 2024-08-27
MAP-1571 revert the removal of gradle property

GurnankCheema:
Removal was done in previous commit and is possibly the cause of a build fail in Circle


marcusaleman (approved):
LGTM

PR #247, opened 2024-08-27
MAP-1457 301 NomisMigrationRequest coverage and DTO responseBody refactor


marcusaleman:
As developer we need to cover Kover report red line scenarios.
This scenario cover Branch % need to be tested.

I have refactor a request in order to reuse it. I would use lambda in the test and do **much abstract** but it would require bit more time from my side. but It would reduce number of line.

There are fewer DeactivatedReason scenarios still need to be covered which I will send in a new PR.

before
image

after
image


Mjwillis (approved):
lgtm

PR #278, opened 2024-10-07
MAP-1619 Provide endpoint to look for a location using prison Id and local name - this will be unique.

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/repository/ResidentialLocationRepository.kt
```diff
@@ -13,7 +13,6 @@ interface ResidentialLocationRepository : JpaRepository<ResidentialLocation, UUI
   fun findAllByPrisonIdAndParentIsNull(prisonId: String): List<ResidentialLocation>
   fun findAllByPrisonIdAndArchivedIsTrue(prisonId: String): List<ResidentialLocation>
   fun findOneByPrisonIdAndPathHierarchy(prisonId: String, pathHierarchy: String): ResidentialLocation?
-  fun findAllByPrisonIdAndActiveIsTrue(prisonId: String): List<ResidentialLocation>
```
jnioche-jd, line 16: are there other parts of the code where this needs removing? or is that already gone but that line had been missed?

Mjwillis: Its not used and I think I never did use it - was just cleaning up

ashleygyngell (approved):

PR #279, opened 2024-10-07
MAP-1666 Don't show non cell locations in the de-activated lists

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -924,7 +924,7 @@ class LocationService(
     }
 
     return (
-      startLocation?.findAllLeafLocations() ?: cellLocationRepository.findAllByPrisonIdAndActive(prisonId, false)
+      startLocation?.cellLocations() ?: cellLocationRepository.findAllByPrisonIdAndActive(prisonId, false)
```
Mjwillis, line 927: This is the fix

jnioche-jd (approved):

PR #288, opened 2024-10-21
MAP-1699 NOMIS local name incorrectly displays AFTER converting residential cell to non-residential room

Mjwillis:


marcusaleman (approved):
LGTM

PR #293, opened 2024-10-23
MAP-1735 Upgrade dependencies

Mjwillis:


marcusaleman (approved):
LGTM

jnioche-jd (changes requested):
Jackson will cause you problem if updated
see https://github.com/ministryofjustice/prison-register/pull/278

jnioche-jd (approved):

PR #302, opened 2024-10-30
MAP-1776 Allow duplicate Local names across different branches, whilst enforcing uniqueness among direct siblings

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResource.kt
```diff
@@ -215,8 +215,11 @@ class LocationResource(
     @Schema(description = "Alternative description to display for location", example = "Wing A", required = false)
     @Size(max = 30, message = "Description must be less than 31 characters")
     @PathVariable localName: String,
+    @Schema(description = "The level above in this hierarchy to look below, empty will check prison level", example = "de91dfa7-821f-4552-a427-bf2f32eafeb0", required = false)
```
Mjwillis, line 218: @ashleygyngell this is the new parameter when checking in the UI for duplicates.

ashleygyngell (approved):

PR #308, opened 2024-11-15
MAP-1840  Force version of netty common to fix security issue + update sqs boot starter to 5.1.1

marcusaleman:
Scanning gave 2 vulnerabilities medium 

This should fix them

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,7 +21,7 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
   implementation("org.springframework.boot:spring-boot-starter-validation")
 
-  implementation("uk.gov.justice.service.hmpps:hmpps-sqs-spring-boot-starter:5.0.1")
+  implementation("uk.gov.justice.service.hmpps:hmpps-sqs-spring-boot-starter:5.1.1")
   implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:2.7.0")
 
   runtimeOnly("org.flywaydb:flyway-database-postgresql")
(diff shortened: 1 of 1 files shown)


GurnankCheema (approved):
LGTM

build.gradle.kts
```diff
@@ -30,6 +30,7 @@ dependencies {
   implementation("com.fasterxml.uuid:java-uuid-generator:5.1.0")
   implementation("org.springdoc:springdoc-openapi-starter-webmvc-ui:2.6.0")
   implementation("com.fasterxml.jackson.module:jackson-module-kotlin:2.17.2")
+  implementation("io.netty:netty-common:4.1.115.Final")
```
Mjwillis, line 33: Are you sure we need to so this.  The SQS library should be fixed not our service

Mjwillis (commented):
Not sure this is needed.  SQS should be fixed

build.gradle.kts
jnioche-jd: @Mjwillis 

We get the dependency from `hmpps-kotlin-spring-boot-starter` which itself depends on 

https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-dependencies/3.3.5

which is the latest available. You are right that we should raise the issue in `hmpps-kotlin-spring-boot-starter` and suggest the patch above so that everyone benefits from it. 

In the short term and until a new version of hmpps-kotlin-spring-boot-starter is available and fixes it, I think we should still apply this small change so that we don't get the alerts reminding us of it all the time.

marcusaleman:
we hold this PR on hold to get more information about SQS STARTER update 
@jnioche-jd 
cc
@Mjwillis  @michalnawrockidj  @GurnankCheema

GurnankCheema (approved):
LGTM

PR #310, opened 2024-11-18
MAP-1849 Use a discriminator for location inheritance

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Location.kt
```diff
@@ -40,7 +40,7 @@ import java.util.*
 import uk.gov.justice.digital.hmpps.locationsinsideprison.dto.Location as LocationDto
 
 @Entity
-@DiscriminatorFormula("case when residential_housing_type IS NULL then 'NON_RESIDENTIAL' when location_type = 'CELL' then 'CELL' when code IN ('RECP', 'COURT', 'TAP', 'CSWAP') then 'VIRTUAL' else 'RESIDENTIAL' end")
+@DiscriminatorColumn(name = "location_type_discriminator")
```
Mjwillis, line 43: Remove the formula and use a new column

src/main/resources/db/migration/V1_39__add_discriminator.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE location ADD COLUMN location_type_discriminator VARCHAR(20);
+
+update location
```
Mjwillis, line 3: This updates the entire database

PR #313, opened 2024-11-19
MAP-1840  This is the upstream fix done for NETTY version.

marcusaleman:
This is the upstream fix done for NETTY version.


Mjwillis (approved):
LGTM

PR #319, opened 2024-11-21
MAP-1870 Filter out usage locations that have a child with same properties

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/LocationService.kt
```diff
@@ -184,13 +184,16 @@ class LocationService(
     sortByLocalName: Boolean = false,
     formatLocalName: Boolean = false,
   ): List<LocationDTO> {
-    val rawResult = nonResidentialLocationRepository.findAllByPrisonIdAndNonResidentialUsages(prisonId, usageType)
+    val filteredByUsage = nonResidentialLocationRepository.findAllByPrisonIdAndNonResidentialUsages(prisonId, usageType)
+    val allLocationIds = filteredByUsage.map { it.id!! }
+    val filteredResults = filteredByUsage
+      .filter { usageLoc -> usageLoc.findSubLocations().map { it.id!! }.intersect(allLocationIds.toSet()).isEmpty() }
```
Mjwillis, line 190: This is the magic bit, it excludes locations that have a child which is returned in the same list that has the same usage type

jimbali (approved):

PR #327, opened 2024-12-17
MAP-1952: Update CONVERTED_CELL_TYPE label from `Converted cell type` to `Non-residential room`.

ashleygyngell:
…e` to `Non-residential room`


Mjwillis (approved):
LGTM

PR #330, opened 2024-12-20
MAP-1954: Modified how changes to cell type add history events. Inste…

ashleygyngell:
…ad of saving each type change individually using the addHistory method, cell types are now saved as a list of values.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationTransformResourceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationTransformResourceTest.kt
@@ -373,6 +373,55 @@ class LocationTransformResourceTest : CommonDataTestBase() {
           )
         }
       }
+
+      @Test
+      fun `can update specialist cell types and record history with changes as a list`() {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationTransformResourceTest.kt
```diff
@@ -373,6 +373,55 @@ class LocationTransformResourceTest : CommonDataTestBase() {
...
+          .expectStatus().isOk
+
+        webTestClient.put().uri("/locations/${cell1.id}/specialist-cell-types")
+          .headers(setAuthorisation(roles = listOf("ROLE_MAINTAIN_LOCATIONS"), scopes = listOf("write")))
+          .header("Content-Type", "application/json")
+          .bodyValue(jsonString(emptySet<SpecialistCellType>()))
+          .exchange()
+          .expectStatus().isOk
+
+        webTestClient.get().uri("/locations/${cell1.id}?includeHistory=true")
```
marcusaleman, line 409: Is there any scenario where includeHistory=true should be empty ?

Mjwillis (approved):
LGTM

PR #332, opened 2025-01-13
MAP-1953 Audit Status and Working Capacity changes when deactivating/reactivating locations

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Location.kt
```diff
@@ -631,6 +631,16 @@ abstract class Location(
         dataChanged = true
       }
 
+      if (this is Cell) {
+        addHistory(
+          LocationAttribute.OPERATIONAL_CAPACITY,
+          getWorkingCapacityIgnoreParent().toString(),
+          "0",
```
Thource, line 638: We've got "0" here but "None" in Cell.kt, is this correct?

Mjwillis: Yes only None when converted from non-cell

PR #347, opened 2025-01-29
MAP-2117 Format proposed reactivation date in location history and related attributes to correct format

Mjwillis:


Thource:
Will frontend receive these dates in dd/MM/yyyy too? if so we may need to make changes there

Mjwillis:
@Thource its just text in UI

Thource (approved):

PR #357, opened 2025-03-21
MAP-2305 integrate DPR library in API with transaction report example

Mjwillis:


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/integration/ResourceSecurityTest.kt
```diff
@@ -20,6 +20,14 @@ class ResourceSecurityTest : SqsIntegrationTestBase() {
     "GET /v3/api-docs/swagger-config",
     " /error",
     "PUT /queue-admin/retry-all-dlqs",
+    "GET /reports/{reportId}/{reportVariantId}/count",
```
Mjwillis, line 23: This is annoying as the DPR endpoints are not secured

PR #369, opened 2025-05-02
HMAI-492 - Queue listener for updates from HMPPS Integration API

tomwatkins1994:
This PR adds a new listener service which will listen to events on the new `update_from_external_system_events_queue` and process them. For now this is just locations that have been temporarily deactivated in an external system.


src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/integration/UpdateFromExternalSystemEventsTest.kt
```diff
@@ -0,0 +1,142 @@
...
+
+  @BeforeEach
+  fun `clear queues`() {
+    queueSqsClient.purgeQueue(PurgeQueueRequest.builder().queueUrl(queueUrl).build())
+    queueSqsDlqClient.purgeQueue(PurgeQueueRequest.builder().queueUrl(dlqUrl).build())
+  }
+
+  @Nested
+  @DisplayName("Location temporarily deactivated event")
+  inner class CreateVisit {
```
Mjwillis, line 58: CreateVisit?

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/integration/UpdateFromExternalSystemEventsTest.kt
```diff
@@ -0,0 +1,142 @@
...
+        verify(
+          locationService,
+          times(0),
+        ).deactivateLocations(any<DeactivateLocationsRequest>())
+      }
+    }
+  }
+
+  @Test
+  fun `will write an invalid visit write event to the dlq`() {
```
Mjwillis, line 125: Visit?

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerServiceTest.kt
```diff
@@ -0,0 +1,134 @@
...
+  private val updateFromExternalSystemListenerService = UpdateFromExternalSystemListenerService(objectMapper, locationService)
+
+  @BeforeEach
+  internal fun setUp() {
+    Mockito.reset(locationService)
+  }
+
+  @Nested
+  @DisplayName("Location temporarily deactivated event")
+  inner class CreateVisit {
```
Mjwillis, line 38: Visit?

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerServiceTest.kt
```diff
@@ -0,0 +1,134 @@
...
+      assertThrows<Exception> {
+        updateFromExternalSystemListenerService.onEventReceived(message)
+      }
+      verify(locationService, times(0)).getLocationByKey(any<String>(), any<Boolean>(), any<Boolean>())
+      verify(locationService, times(0)).deactivateLocations(any<DeactivateLocationsRequest>())
+    }
+  }
+
+  @Test
+  fun `will throw an an exception when invalid visit write event passed in`() {
```
Mjwillis, line 117: visit?

Mjwillis (commented):
Will discuss with @garethmdavies about this - as stated yesterday, I don't think we should be using this approach. 

The calling service should be handling calls to the API, error, retries etc and not making this service deal with these issues.  But it this is the strategic approach we are adopting then I will accept.

Some thoughts:
- Currently locations are tightly controlled through the front end with validation and other checks, although the API also applies most of these, the queue feels like a back door to changing important data.
- Clients, how many queues?
- Responsibility for handling failures
- Alerting and monitoring
- Validation, 
- Misuse of the turning off locations
-

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerService.kt
```diff
@@ -0,0 +1,51 @@
...
+          val key = temporarilyDeactivateLocationByKeyRequest.key
+          val location = locationService.getLocationByKey(key) ?: throw Exception("No location found for key: $key")
+          val temporaryDeactivationLocationRequest = TemporaryDeactivationLocationRequest(
+            deactivationReason = temporarilyDeactivateLocationByKeyRequest.deactivationReason,
+            deactivationReasonDescription = temporarilyDeactivateLocationByKeyRequest.deactivationReasonDescription,
+            proposedReactivationDate = temporarilyDeactivateLocationByKeyRequest.proposedReactivationDate,
+            planetFmReference = temporarilyDeactivateLocationByKeyRequest.planetFmReference,
+          )
+          locationService.deactivateLocations(DeactivateLocationsRequest(mapOf(location.id to temporaryDeactivationLocationRequest)))
+          LOG.info("Location temporarily deactivated: $key (${location.id})")
```
Mjwillis, line 42: That will already be logged in the service call

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerService.kt
```diff
@@ -0,0 +1,51 @@
...
+          val temporarilyDeactivateLocationByKeyRequest = sqsMessage.toTemporarilyDeactivateLocationByKeyRequest()
+          val key = temporarilyDeactivateLocationByKeyRequest.key
+          val location = locationService.getLocationByKey(key) ?: throw Exception("No location found for key: $key")
+          val temporaryDeactivationLocationRequest = TemporaryDeactivationLocationRequest(
+            deactivationReason = temporarilyDeactivateLocationByKeyRequest.deactivationReason,
+            deactivationReasonDescription = temporarilyDeactivateLocationByKeyRequest.deactivationReasonDescription,
+            proposedReactivationDate = temporarilyDeactivateLocationByKeyRequest.proposedReactivationDate,
+            planetFmReference = temporarilyDeactivateLocationByKeyRequest.planetFmReference,
+          )
+          locationService.deactivateLocations(DeactivateLocationsRequest(mapOf(location.id to temporaryDeactivationLocationRequest)))
```
Mjwillis, line 41: No events are being raised on completion of the deactivation this will mean that NOMIS is out of sync!  You need to add the code to raised the correct events for the correct location - if its a parent location then the events cascade downwards

Mjwillis (changes requested):
Need to add events so that NOMIS will sync

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerService.kt
```diff
@@ -0,0 +1,53 @@
...
+          val temporaryDeactivationLocationRequest = event.let {
+            TemporaryDeactivationLocationRequest(
+              deactivationReason = it.deactivationReason,
+              deactivationReasonDescription = it.deactivationReasonDescription,
+              proposedReactivationDate = it.proposedReactivationDate,
+              planetFmReference = it.planetFmReference,
+            )
+          }
+          val deactivateLocationsRequest = DeactivateLocationsRequest(mapOf(event.id to temporaryDeactivationLocationRequest))
+          val result = locationService.deactivateLocations(deactivateLocationsRequest)
```
Mjwillis, line 43: redundant variable

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerService.kt
```diff
@@ -0,0 +1,53 @@
...
+          deactivate(result)
+        }
+        else -> throw Exception("Cannot process event of type ${sqsMessage.eventType}")
```
Mjwillis, line 46: This will go back on the DLQ. What happens then?  I assume we can't process it? So clear it down?

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerService.kt
```diff
@@ -0,0 +1,53 @@
...
+      }
+    } catch (e: Exception) {
```
Mjwillis, line 48: Not sure this extra catch does anything, the exception would be logged and the message do back on the DLQ

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerService.kt
```diff
@@ -0,0 +1,53 @@
...
+import uk.gov.justice.digital.hmpps.locationsinsideprison.resource.DeactivateLocationsRequest
+import uk.gov.justice.digital.hmpps.locationsinsideprison.resource.EventBaseResource
+
+const val UPDATE_FROM_EXTERNAL_SYSTEM_QUEUE_CONFIG_KEY = "updatefromexternalsystemevents"
+
+@Service
+class UpdateFromExternalSystemListenerService(
+  private val objectMapper: ObjectMapper,
+  private val locationService: LocationService,
+) : EventBaseResource() {
```
Mjwillis, line 19: I think now this super class needs  a name without `Resource` in it

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/integration/UpdateFromExternalSystemEventsTest.kt
```diff
@@ -0,0 +1,151 @@
...
+        landingN1.cellLocations().map { it.getPathHierarchy() },
+        false, // Return no results as this will cause deactivate to fail
+      )
+
+      val messageId = UUID.randomUUID().toString()
+      val updateFromExternalSystemEvent = UpdateFromExternalSystemEvent(
+        messageId = messageId,
+        eventType = "LocationTemporarilyDeactivated",
+        messageAttributes = mapOf(
+          "id" to landingN1.id,
```
Mjwillis, line 84: This is a landing.  - are you allowing clients to deactivate whole landings?  This will cascade down to all the cells.  Not sure we want to allow that!

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/integration/UpdateFromExternalSystemEventsTest.kt
```diff
@@ -0,0 +1,151 @@
...
+      queueSqsClient.sendMessage(
+        SendMessageRequest.builder().queueUrl(queueUrl).messageBody(message).build(),
+      )
+
+      await untilCallTo { getNumberOfMessagesCurrentlyOnQueue() } matches { it == 1 }
+      await untilCallTo { getNumberOfMessagesCurrentlyOnQueue() } matches { it == 0 }
+      await untilCallTo { getNumberOfMessagesCurrentlyOnDlq() } matches { it == 0 }
+
+      await untilAsserted { verify(locationService, times(1)).deactivateLocations(any<DeactivateLocationsRequest>()) }
+    }
```
Mjwillis, line 101: Need to assert that the UPDATE event is sent: 
``` kotlin
    getDomainEvents(2).let {
          assertThat(it.map { message -> message.eventType to message.additionalInformation?.key }).containsExactlyInAnyOrder(
            "location.inside.prison.deactivated" to "NMI-A-1-001",
             "location.inside.prison.deactivated" to "NMI-A-1",
          )
        }
```

Mjwillis (changes requested):
- Don't allow parent deactivation unless there is a clear reason this should be allowed
- Need to assert that events are coming through
- couple of other minor comments

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerService.kt
tomwatkins1994: At that point it's more for visibility, there won't be a way to fix it but so either clear it off manually or wait for the retention policy to remove it. There shouldn't really be a case where this happens anyway as the event name is not a user input.

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/integration/UpdateFromExternalSystemEventsTest.kt
tomwatkins1994: As far as I know it's likely our use case should be cells only, but I will double check and apply validation at our end if so. Would you be happy with that validation living only on our API or do you want it in both places?

tomwatkins1994: We have applied validation on our end for cell only and also replicated this in this repo

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/UpdateFromExternalSystemListenerService.kt
tomwatkins1994: I've renamed it to `EventBase`. I left it in the `resource` folder as there wasn't an obvious place to move this to. Any ideas?

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/integration/UpdateFromExternalSystemEventsTest.kt
tomwatkins1994: This is done, just needed to also assert that the amend event was sent for the parent locations

Mjwillis (approved):
Approved

PR #377, opened 2025-05-16
MAP-2411 return all child locations deleted - but don't publish if DRAFT

Mjwillis:
### Summary
This Pull Request introduces changes to the LocationService class to return all the sub location that are affect  by the delete and checks that no events are emitted for draft location that were deleted.
- LocationService.kt:
  - Modified the toDto() method call to include children by passing a new argument includeChildren = true.
  - This change affects how DTOs are generated, potentially adding more detailed data.
- DraftLocationResourceTest.kt:
  - Added assertions in existing tests to check that the number of messages in the queue is zero after certain operations.
  - These assertions enhance the test coverage by validating message queue states.


ashleygyngell (approved):
👍

PR #395, opened 2025-06-05
MAP-2471 Only return all the location data when looking at an individual certificate or the current one

Mjwillis:
This Pull Request introduces changes to the handling of locations in CellCertificateDto to make them optional, adds a parameter to include or exclude these locations in service methods, and updates corresponding tests to validate the new behavior.
### Main Changes
- CellCertificateDto.kt:
  - Made locations field optional (List changed to List?) with a default value of null.
  - Modified the from method to accept a showLocations parameter. Based on this parameter, either include filtered locations or set locations to null.
- CellCertificateService.kt:
  - Updated service methods (getCellCertificate and getCurrentCellCertificateForPrison) to pass showLocations = true to ensure locations are included when needed.
- CellCertificateResourceTest.kt:
  - Added new test validations to assert locations presence or absence in API responses based on specific cases.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -20,7 +20,7 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
   implementation("org.springframework.boot:spring-boot-starter-validation")
-  implementation("uk.gov.justice.service.hmpps:hmpps-digital-prison-reporting-lib:8.3.3")
+  implementation("uk.gov.justice.service.hmpps:hmpps-digital-prison-reporting-lib:8.3.4")
   implementation("uk.gov.justice.service.hmpps:hmpps-sqs-spring-boot-starter:5.4.5")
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/CellCertificateService.kt
```diff
@@ -58,7 +58,7 @@ class CellCertificateService(
   fun getCellCertificate(id: UUID): CellCertificateDto {
     val cellCertificate = cellCertificateRepository.findById(id)
       .orElseThrow { CellCertificateNotFoundException(id) }
-    return CellCertificateDto.from(cellCertificate)
+    return CellCertificateDto.from(cellCertificate, showLocations = true)
```
Mjwillis, line 61: Only return all the location data when looking at an individual certificate or the current one

dafhoustonmoj (approved):
LGTM

PR #401, opened 2025-06-11
MAP-2462 Refactor DTOs and entities; streamline certification and location handling

Mjwillis:


src/main/resources/db/migration/V1_62__back_links_nullable.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE certification_approval_request_location ALTER COLUMN certification_approval_request_id DROP NOT NULL;
```
Mjwillis, line 1: As I've remove the backlinks from the entities I've had to make the FK nullable due to the way JPA works. The column we actually never be null but within the transaction it will be briefly as JPA issues an insert and then an update

PR #443, opened 2025-09-16
MAP-2778 Add fallback to default next level type for sub-location descriptions in ResidentialSummary

Mjwillis:
### Summary
This pull request adds a fallback mechanism to default to the next level type for sub-location descriptions within the `ResidentialSummary` component.

### Testing
- [ ] Verify that sub-location descriptions correctly use the default next level type when none is specified.


dafhoustonmoj (approved):
LGTM

PR #447, opened 2025-09-17
SDIT-3040 New location field needs adding to sync

steverendell:
Allow null in all the patch request endpoint fields


Mjwillis (approved):
Lgtm - might be nice to do it the way mike h suggested but that's a future change

PR #627, opened 2026-04-24
MAP-3187 Fix cell reactivation skipped when reactivating a single cell

Mjwillis:
## Summary

- Fixed a bug where reactivating a single cell (rather than a wing or landing with sub-locations) would silently skip the reactivation — the approval location was never marked for reactivation because the existing logic only iterated over `subLocations`, which is empty for a leaf cell
- The top-level approval location is now processed directly before iterating sub-locations, ensuring a single-cell approval request correctly marks the cell for reactivation and updates its capacity details
- Fixed a related issue where `cellReactivationChanges` referencing the top-level location itself was not handled, only child locations were looked up
- Added `locationRepository.saveAll()` in `SharedLocationService` to ensure reactivated locations are persisted; also annotated the class with `@Transactional`
- Exposed `reactivateThisLocation` in `CertificationApprovalRequestLocationDto` so the flag is visible in the API response
- Extracted `updateLocationCapacitiesForApproval` and `updateApprovalLocationDetails` helpers in `ApprovalRequestService` to eliminate duplication and consistently handle both top-level and sub-locations

## Test plan

- [ ] New integration test `can request to reactivate a cell and then approve` covers the single-cell reactivation path end-to-end, verifying the approval request, approval, certificate state, and domain events
- [ ] New integration test `can deactivate a cell and request approval` covers single-cell deactivation with approval
- [ ] Existing wing-level cascade and explicit reactivation tests renamed and updated to use the new shared `deactivateLocation` helper, confirming no regression

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/CertificationApprovalRequestLocationDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/CertificationApprovalRequestLocationDto.kt
@@ -69,6 +69,9 @@ data class CertificationApprovalRequestLocationDto(
   @param:Schema(description = "Converted cell type", example = "OFFICE", required = false)
   val convertedCellType: ConvertedCellType? = null,
 
+  @param:Schema(description = "Indicates this location will be reactivated", example = "true", required = false)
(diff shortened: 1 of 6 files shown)


dafhoustonmoj (approved):
LGTM

PR #628, opened 2026-04-27
MAP-3188 Add `inactiveStatus` to support inactive cells screen categorisation

Mjwillis:
### Summary

This PR adds a new `inactiveStatus` field to the `LocationDTO` to allow the inactive cells screen to categorise temporarily deactivated cells into three distinct groups. This supports the UI requirement to display inactive cells in separate sections based on their relationship to the official cell certificate.

### Background

The inactive cells screen (supported by `GET /prison/{prisonId}/inactive-cells`) previously returned no information about *why* a cell is inactive relative to its certificate. The screen needs to divide inactive cells into three sections:

1. **Temporary inactive cells** — inactive but the certificate has not been updated to reflect 0 working capacity
2. **Inactive cells with certificate change requests** — inactive with a pending deactivation approval request
3. **Inactive cells with capacity decreased on cell certificate** — the working capacity on the certificate has already been reduced to match the inactive state

#### `LocationDto.kt`
- Added a new `inactiveStatus: InactiveStatus?` field to the `Location` DTO. This is `null` when the location is active or when no cell certificate is available (i.e. for prisons that do not require certification approval).
- Added new `InactiveStatus` enum with three values:
    - `INACTIVE_TEMP` — cell is inactive but the certificate working capacity does not yet reflect 0
    - `INACTIVE_PEND_CHANGE_REQ` — cell has a pending deactivation certification approval request
    - `INACTIVE_MATCHING_CELL_CERT` — cell's working capacity matches the cell certificate (i.e. certificate has been updated to 0)

#### `Cell.kt`
- Added `deriveInactiveStatus(cellCertificateLocation)` private method on `Cell` to calculate the appropriate `InactiveStatus`:
    - If the cell has a pending certification approval → `INACTIVE_PEND_CHANGE_REQ`
    - Else if the cell certificate working capacity matches the calculated working capacity → `INACTIVE_MATCHING_CELL_CERT`
    - Otherwise → `INACTIVE_TEMP`
- The `toDto()` method now maps `cellCertificateLocation` to `inactiveStatus` via this method.

#### `LocationService.kt`
- `getResidentialInactiveLocations`: Now checks `isCertificationApprovalRequired` once per request (avoiding repeated calls per cell), and conditionally fetches the cell certificate per location only when required. If certification approval is not required for the prison, the DB lookup is skipped entirely and `inactiveStatus` will be `null`.
- `temporarilyDeactivateLocations`: Extracted `certificationApprovalRequired` as a local variable (previously called inline) and passes the cell certificate to `toDto()` so deactivated locations also carry their `inactiveStatus` in the response.

#### `CertificationApprovalResourceTest.kt`
- Added assertions across all three deactivation approval scenarios (pending, approved, rejected) to verify the correct `inactiveStatus` is returned both on individual location responses and via the `inactive-cells` list endpoint.
- Refactored a duplicated inline deactivation block to use the shared `deactivateLocation()` helper.
- Corrected a minor pre-existing test data typo (`"The cell as been flooded"` → `"The wing is being renovated"`) and removed redundant duplicate request steps.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationDto.kt
@@ -133,6 +133,9 @@ data class Location(
   @param:Schema(description = "Indicates the location is enabled", example = "true", required = true, deprecated = true)
   val active: Boolean = true,
 
+  @param:Schema(description = "If this location is inactive, indicates the type of inactive status", example = "true", required = true, deprecated = true)
+  val inactiveStatus: InactiveStatus? = null,
+
   @param:Schema(description = "In-cell sanitation", required = false, example = "true")
   val inCellSanitation: Boolean? = null,
 
@@ -259,6 +262,14 @@ data class Location(
     result = 31 * result + pathHierarchy.hashCode()
     return result
   }
+}
+
+enum class InactiveStatus(
+  val description: String,
+) {
+  INACTIVE_TEMP("Temporary inactive cell"),
+  INACTIVE_PEND_CHANGE_REQ("Inactive cell with certificate change requests"),
+  INACTIVE_MATCHING_CELL_CERT("Inactive cell with capacity decreased on cell certificate"),
 }
 
 @Schema(description = "Pending changes")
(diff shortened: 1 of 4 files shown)


dafhoustonmoj (approved):
LGTM

PR #648, opened 2026-05-15
fix: Allow used for update without cell cert change approval

jimbali:
Updates to the usedFor value of a cell should not require a cell cert change request.

MAP-3204

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ sonar-project.properties
 
 #Helm
 **/Chart.lock
+
+#Synk
+.snyk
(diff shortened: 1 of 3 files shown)


.snyk
```diff
@@ -0,0 +1,19 @@
+# WARNING - THIS FILE WAS GENERATED BY THE hmpps-gradle-spring-boot GRADLE PLUGIN
```
Mjwillis: I think add  to git ignore

Mjwillis (changes requested):

Mjwillis (approved):
LGTM

PR #660, opened 2026-05-29
MAPA-104: Allow converting a cell to a non-residential room via the certification-approval workflow 

Mjwillis:
## Summary

Converting a cell to a non-residential room (`PUT /locations/{id}/convert-cell-to-non-res-cell`) was
previously **rejected outright** when the prison operates with certification approval enabled
(`certificationApprovalRequired = true`) — the service threw
`ChangesCannotBeMadeWithoutCertificationApprovalException`.

This change allows the conversion on cert-approval prisons by routing it through the existing approval
workflow (the same machinery used for capacity changes, specialist-cell-type changes, deactivation, etc.),
following the pattern established by MAPA-103.

There is **no new endpoint**. Requesting the conversion continues to use the existing convert endpoint;
approve / reject / withdraw use the existing generic
`PUT /certification/location/{approve,reject,withdraw}` endpoints. Prisons that do **not** require
certification approval are unaffected — they convert directly as before.

### Prison requires certification approval

| Stage | Outcome |
|---|---|
| **Request** | Checks no prisoners are present (the cell is about to be deactivated). The cell is **temporarily deactivated** (`status = INACTIVE`, `temporarilyOffCellCert = true`) with the new deactivation reason **"Convert cell to non-residential room"**, and a `ConvertToNonResidentialCellApprovalRequest` is raised capturing the cell's current state. A `LOCATION_DEACTIVATED` domain event is published. While pending, the cell appears as `LOCKED_INACTIVE` (inactive + change requested) and remains on the existing certificate at its current certified values. |
| **Approve** | The temporary deactivation is cleared and the conversion is applied, so the cell becomes an **active, non-residential converted room** (capacity removed, de-certified, specialist cell types and used-for cleared). A new certificate is generated reflecting the non-residential cell. A `LOCATION_AMENDED` domain event is published. **No** prisoner check is performed on approval (the cell is already inactive, so no prisoners can be present). |
| **Reject / Withdraw** | The cell **remains temporarily inactive** with the same reason. No conversion takes place and the certificate is unchanged. |

### Prison does not require certification approval

Unchanged: the cell is converted immediately and a `LOCATION_AMENDED` event is published.

## How current vs. new values are presented

The approval request captures the cell's **current** capacity (CNA / working / max), specialist cell types
and sanitation so the UI can render the change as `current → new`. Because the conversion removes all of
these, the **new (pending)** values are deliberately left `null` so the UI shows, for example:

- `Working capacity 2 → None`
- `CNA 1 → None`
- `Sanitation Yes → None`
- `Specialist cell types [...] → None`

The only value with a populated "new" side is the converted cell type, e.g. `None → OFFICE`.

Current values are sourced exactly as the other approval types do them:

- Current capacity from the approval request's location snapshot (`currentWorkingCapacity` etc.) plus the
  `*Change` deltas via `refreshCapacities()`.
- Current specialist cell types from the snapshot (`getTopLevelLocation().getSpecialistCellTypesFromList()`).
- Current sanitation stored on the request entity as `currentInCellSanitation` (reusing the existing column,
  as `SanitationChangeApprovalRequest` does).

### New `DeactivatedReason` value
`CONVERT_CELL_TO_ROOM` ("Convert cell to non-residential room", sequence 10). Stored as a string in existing
columns — no migration required for the enum value itself.

### New `ApprovalType` value
`CONVERT_CELL_TO_ROOM` with `hasPendingValues = false` (consistent with `DEACTIVATION`). The cell is
deactivated immediately on request and the snapshot captures current values with the working-capacity change
applied via the same fix-up `deactivateLocations` uses — so pending values must not be folded into the
hierarchy snapshot.

### New entity `ConvertToNonResidentialCellApprovalRequest`
- `@DiscriminatorValue("CONVERT_CELL_TO_ROOM")`, extends `LocationCertificationApprovalRequest`.
- Stores `convertedCellType` / `otherConvertedCellType` (new columns) and `currentInCellSanitation`
  (reuses the existing `current_in_cell_sanitation` column — same un-named logical mapping as
  `SanitationChangeApprovalRequest` so Hibernate treats it as the same physical column).
- `toDto()` surfaces `convertedCellType`, `otherConvertedCellType`, `currentSpecialistCellTypes` and
  `currentInCellSanitation`, leaving the pending capacity / specialist / sanitation fields null.
- Does **not** override `approve` / `reject` / `withdraw` — the cell was deactivated at request time and must
  stay as-is on reject/withdraw, which the base behaviour already does.

### Flyway migration `V1_90__convert_cell_to_room_approval.sql`
Adds `converted_cell_type VARCHAR(60)` and `other_converted_cell_type VARCHAR(255)` to
`certification_approval_request`. All other columns already exist.

### `Cell`
- Split `convertToNonResidentialCell` and `updateNonResidentialCellType` into a public method (which keeps the
  `hasPendingCertificationApproval()` guard) and an `internal` `apply…` method holding the body. The guard-free
  variants are used when applying an already-approved conversion (the request is still `PENDING` at that point,
  so the public guard would otherwise throw).
- Added `clearTemporaryDeactivationForConversion(...)` — clears the temporary-deactivation status fields
  (status → ACTIVE, `temporarilyOffCellCert = false`, reason/date/etc. nulled) **without** re-certifying or
  restoring capacity (the conversion that follows removes them anyway).
- Added `requestApprovalForConvertToNonResidentialCell(...)`.

### `LocationService.convertToNonResidentialCell`
Branches on `certificationApprovalRequired`. When required, delegates to a new private
`requestConversionApproval(...)` which performs the prisoner check, raises the approval request, applies the
snapshot capacity fix-up + `refreshCapacities()`, temporarily deactivates the cell, and returns the now-inactive
location. The original (non-approval) path is unchanged.

### `ApprovalDecisionService`
New `is ConvertToNonResidentialCellApprovalRequest` branch in `handleComplexApprovalProcesses`. On approval it
clears the temporary deactivation, applies the (guard-free) conversion, and returns a `LOCATION_AMENDED` event
map. Runs before the request is marked approved and before the new certificate is built, so the certificate
reflects the converted cell.

### `LocationResidentialResource`
The convert endpoint now selects the domain event after the service call: `LOCATION_DEACTIVATED` when an
approval was raised (`pendingApprovalRequestId != null`), otherwise `LOCATION_AMENDED`. The single-`Location`
response shape is unchanged.

### `CertificationApprovalRequestDto`
Added optional `convertedCellType` and `otherConvertedCellType` fields.

## Tests

New `@Nested` class `ConvertCellToNonResidentialRoomTest` in `CertificationApprovalResourceTest`:

- **Request** temporarily deactivates the cell, raises a `CONVERT_CELL_TO_ROOM` approval, captures current
  values with new values null (and `convertedCellType = OFFICE`), and publishes `LOCATION_DEACTIVATED`
  (+ parent `LOCATION_AMENDED`).
- **Request** captures current specialist cell types.
- **Request with a prisoner present** returns `409` (`DeactivationErrorLocationsContainPrisoners`) and raises
  no approval.
- **Approve** converts the cell to an active non-residential room (capacity removed, de-certified, specialist
  types cleared), regenerates the certificate (totals drop accordingly), and publishes `LOCATION_AMENDED`.
- **Reject** leaves the cell temporarily inactive with the conversion reason; certificate unchanged.
- **Withdraw** leaves the cell temporarily inactive.
- **Non-approval prison** converts directly (regression).

`LocationConstantsIntTest` updated to include the new deactivation reason in the
`/constants/deactivated-reason` response.

## Test plan

- [x] `./gradlew ktlintCheck` — clean
- [x] `./gradlew test` — all 790 tests pass
- [ ] Manual check: on a cert-approval prison, request a conversion → cell becomes inactive (off cert),
  approval captures current capacity / specialist types / sanitation with new = None
- [ ] Manual check: approve → cell becomes an active non-residential room and the new certificate reflects it
- [ ] Manual check: reject / withdraw → cell stays inactive with the "Convert cell to non-residential room"
  reason and the certificate is unchanged
- [ ] Manual check: non-cert-approval prison still converts directly


dafhoustonmoj (approved):
LGTM

PR #668, opened 2026-06-04
MAPA-110 Fix missing specialistCellTypes / currentSpecialistCellTypes on certification approval request locations

Mjwillis:
## Summary

When fetching a certification approval request via `GET /certification/request-approvals/{id}`, the
entries in the `locations` array were missing `specialistCellTypes` and `currentSpecialistCellTypes`,
even though those fields were correctly populated on the top-level approval object.

This was most visible after creating a **SPECIALIST_CELL_TYPE** approval request to add a specialist
cell type to a cell that had none: the top-level `specialistCellTypes` showed the proposed types, but
the location inside the array showed neither field. Both fields are dropped from the JSON by
`@JsonInclude(NON_NULL)` when null, so they simply disappeared from the response.

## Root cause

In `ResidentialLocation.toCertificationApprovalRequestLocation()` (used to build the `locations`
hierarchy on an approval request):

1. **`currentSpecialistCellTypes` was never set** on `CertificationApprovalRequestLocation` — the
   constructor call omitted it entirely, so it stayed `null`.
2. **`specialistCellTypes` ignored the pending change** — it was populated from
   `getSpecialistCellTypesAsCSV()`, which reads the *live* cell state. For the "add a type to a cell
   that currently has none" case this resolved to empty → `null`, so the proposed types never appeared
   on the location.

A secondary coupling: `SpecialistCellTypeChangeApprovalRequest.toDto()` derived the **top-level**
`currentSpecialistCellTypes` from the location entity's `specialistCellTypes` field. Once that field
was corrected to hold pending values, the top-level value had to be repointed to read the new
`currentSpecialistCellTypes` field instead.

## Changes

**`ResidentialLocation.kt`**
- `toCertificationApprovalRequestLocation()` now sets:
    - `currentSpecialistCellTypes = getSpecialistCellTypesAsCSV()` (live state), and
    - `specialistCellTypes = getSpecialistCellTypesAsCSV(includePending = includeDraftOrPending)`
      (pending-aware).
- Added a `getSpecialistCellTypesAsCSV(includePending: Boolean)` overload that aggregates child cells
  via the existing pending-aware `Cell.getSpecialistCellTypesForCell(includePending)`. The no-arg
  version is unchanged, so the other callers (`CellCertificateService`, `ApprovalRequestService`, and
  the certificate-location builder) are unaffected.

**`CertificationApprovalRequestLocation.kt`**
- Added `getCurrentSpecialistCellTypesFromList()`, mirroring the existing
  `getSpecialistCellTypesFromList()`, to parse the `currentSpecialistCellTypes` column.

**`SpecialistCellTypeChangeApprovalRequest.kt`**
- Repointed the top-level `currentSpecialistCellTypes` to use
  `getCurrentSpecialistCellTypesFromList()` so it keeps reflecting the cell's pre-change state.

## Tests

`CertificationApprovalResourceTest$SpecialistCellTypeTest`:
- Extended `can request approval to add a capacity-affecting specialist cell type` to assert the
  array entry surfaces the proposed `specialistCellTypes` and that `currentSpecialistCellTypes` is
  empty (the exact reported scenario — adding a type to a cell with none).
- Added `request approval surfaces existing specialist cell types as currentSpecialistCellTypes on the
  location`, which gives the cell an existing type, requests a change, and asserts the array entry
  reports `currentSpecialistCellTypes` (old) distinctly from `specialistCellTypes` (proposed). This
  guards the half of the bug where `currentSpecialistCellTypes` was never populated, and fails on the
  pre-fix code.

## Verification

- Full suite: 815 tests pass.
- `ktlintCheck`: clean.


GurnankCheema (approved):
LGTM

PR #674, opened 2026-06-10
MAPA-112 Prevent duplicate cell certificates from concurrent upload processing

Mjwillis:
### Problem

Uploading a large prison (e.g. MKI, 1421 rows, ~8 min to process) produced **3 cell certificates** instead of 1.

The async upload is processed by an `@SqsListener`, and the SQS message is only deleted when processing returns. The `update_cell_certificate_queue` has a **120s visibility timeout** and `maxReceiveCount = 3`, so a multi-minute job let the message become visible again and be **redelivered to other pods**, processing the same upload concurrently — one certificate per run (3 deliveries → 3 certificates).

The existing guards weren't enough: `startProcessing()` only skipped `FINISHED` uploads (an in-flight `STARTED` upload was processed again), and `finish()` used a non-atomic `if (status == FINISHED) return` across separate transactions.

### Change

- New `CellCertificateUploadRepository.findByIdForUpdate` with `@Lock(PESSIMISTIC_WRITE)` (`SELECT … FOR UPDATE`).
- `startProcessing()` claims the upload under the row lock: the first consumer moves `PENDING → STARTED` and commits; concurrent consumers then read `STARTED` and **skip**, so only one run proceeds and a single certificate is created. A 30-minute stale-claim threshold still lets a redelivery re-claim an upload whose consumer crashed mid-processing.
- `finish()` re-locks the row so the `FINISHED` check-and-set is atomic (defence in depth).

The lock is held only for the brief claim/finish transactions — the long per-row loop runs outside it.

### Testing

- New integration test processes the same upload from 3 threads simultaneously and asserts exactly one certificate is produced.
- Full suite: `./gradlew test` — 831 tests pass. `ktlintCheck` clean.

### Follow-ups (separate)

- Raise the queue `visibility_timeout_seconds` in cloud-platform-environments to reduce redelivery churn.
- Reconcile the duplicate MKI certificates already created.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/repository/CellCertificateUploadRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/repository/CellCertificateUploadRepository.kt
@@ -1,6 +1,10 @@
 package uk.gov.justice.digital.hmpps.locationsinsideprison.jpa.repository
 
+import jakarta.persistence.LockModeType
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Lock
+import org.springframework.data.jpa.repository.Query
+import org.springframework.data.repository.query.Param
 import org.springframework.stereotype.Repository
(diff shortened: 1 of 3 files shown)


dafhoustonmoj (approved):
LGTM

PR #675, opened 2026-06-10
MAPA-112 Cell certificate upload: live progress counts and treat location-not-found as a failure

Mjwillis:
## MAPA-112

Two refinements to asynchronous cell certificate upload processing, prompted by the large-prison (MKI) upload.

### 1. Live progress counts

The `processed / skipped / failed` counts on the upload were only written in `finish()`, so a UI refresh showed `0 / 0 / 0` until the whole (multi-minute) job completed.

Each row is already processed in its own `REQUIRES_NEW` transaction, so the counts are now incremented atomically per row (`@Modifying` update committed with the row outcome) and become visible to the read-only GET endpoint immediately — the "so far" totals climb as the upload runs. `finish()` still recomputes the authoritative totals from the location rows (self-heals any drift and the stale-reclaim path).

No API/DTO change: `CellCertificateUploadDto` already exposes these counts.

### 2. "Location not found" is now a failure

A row whose location can't be found during processing was previously `markSkipped("Location not found")`; it is now `markFailed`, so it surfaces in the **Failed** total rather than being quietly skipped. Archived locations and no-change rows remain skips.

### Testing

- New tests: a not-found row ends `FAILED` (`failedRecords` counted), and the atomic increment methods bump the right columns.
- Full suite: `./gradlew test` — 833 tests pass. `ktlintCheck` clean.


dafhoustonmoj (approved):
LGTM

PR #676, opened 2026-06-11
MAPA-127 Record working capacity change in history on cell reactivation

Mjwillis:
# MAPA-127 Record working capacity change in history on cell reactivation

## Problem

When a cell is temporarily deactivated, the service writes a `LocationHistory`
entry recording the working capacity dropping to `0`, but it does **not** change
the cell's stored `Capacity.workingCapacity`.

On approval-driven reactivation (`ApprovalDecisionService.handleReactivation` →
`buildReactivationDetail`) a `Capacity` is always supplied, so in
`Location.reactivate()`:

- `capacityAdjusted` is `true`, which skipped the working-capacity history
  fallback (`if (!capacityAdjusted)`), and
- it relied on `setCapacity()` to log the change — but `setCapacity()` compares
  the new value against the **stored** working capacity. Because deactivation
  never altered the stored value, reactivating to the same value (the normal
  case) made `setCapacity()` a no-op.

The result was an asymmetric audit trail: history showed working capacity
`X → 0` on deactivation but nothing showing it returning to `X` on
reactivation.

## Change

In `Location.reactivate()`, the working-capacity history entry is now recorded
on every cell reactivation (the `!capacityAdjusted` guard has been removed),
using the deactivated value (`previousWorkingCapacity`, `0` while inactive) as
the old value and the now-active working capacity as the new value. The amended
date is aligned with `linkedTransaction.txStartTime` to match the sibling
reactivation history entries.

`addHistory()` already de-duplicates identical entries and is a no-op when the
old and new values match, so genuine capacity changes continue to record
correctly without producing duplicate entries.

## Testing

- Added integration test
  `reactivating a cell via approval records the working capacity returning to its previous value in history`
  to `CertificationApprovalResourceTest`. It deactivates a cell, reactivates it
  via the approval flow back to the same stored capacity, and asserts the
  `changeHistory` contains both the deactivation `1 → 0` entry and the
  reactivation `0 → 1` entry. This test fails before the fix and passes after.
- Regression suites pass: `CertificationApprovalResourceTest`,
  `LocationResourceIntTest`, `LocationCapacityResourceIntTest`,
  `CellCertificateResourceTest`, `LocationNonResidentialResourceTest`, and all
  `jpa.*` tests.
- `./gradlew ktlintCheck` passes.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Location.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/Location.kt
@@ -1077,16 +1077,18 @@ abstract class Location(
           this.residentialHousingType = this.accommodationType.mapToResidentialHousingType()
         }
 
-        if (!capacityAdjusted) {
-          addHistory(
-            LocationAttribute.WORKING_CAPACITY,
-            previousWorkingCapacity.toString(),
-            getWorkingCapacityIgnoreParent().toString(),
-            userOrSystemInContext,
-            LocalDateTime.now(clock),
-            linkedTransaction,
-          )
-        }
+        // Always record the working capacity returning to its active value. Temporary deactivation records the
+        // working capacity dropping to 0 without changing the stored Capacity, so when a cell is reactivated to the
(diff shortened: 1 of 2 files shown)


GurnankCheema (approved):
LGTM

PR #680, opened 2026-06-19
MAPA-131 Accept reason for change on cell to non-residential conversion

Mjwillis:
## What

The **Convert cell to non-residential room** page lets the user "Explain the reason for this change", but `PUT /locations/{id}/convert-cell-to-non-res-cell` did not accept a `reasonForChange`, so the text was silently dropped.

This adds an optional `reasonForChange` to the convert request contract and persists it on the certification-approval flow.

## Changes

- **Contract**: `ConvertCellToNonResidentialLocationRequest` gains an optional `reasonForChange`.
- **Service**: `convertToNonResidentialCell` forwards the reason to `requestConversionApproval`, which now:
  - requires a non-blank reason on prisons that require certification approval (400 `ApprovalRequestRequiresReasonForChange`, matching `cell-mark-change` / `cell-sanitation-change`), and
  - stores it on the `ConvertToNonResidentialCellApprovalRequest` (replacing the previously hardcoded `null`).
- No entity/DB migration needed — the approval request already has the `reasonForChange` column.
- Prisons that do not require approval accept the field but do not persist it.

## Tests

- Approval flow: asserts the supplied reason is stored on the approval request; new test that an approval-required prison rejects a conversion with no reason (and raises no pending request).
- Updated the prisoner-present test to supply a reason so it still exercises the 409 conflict.
- Non-approval happy path now sends a reason, confirming the field is accepted.

All 191 tests in the affected resource test classes pass.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResidentialResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationResidentialResource.kt
@@ -464,7 +464,7 @@ class LocationResidentialResource(
     convertCellToNonResidentialLocationRequest: ConvertCellToNonResidentialLocationRequest,
   ): Location {
     val updatedLocation = with(convertCellToNonResidentialLocationRequest) {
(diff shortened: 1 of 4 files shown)


dafhoustonmoj (approved):
LGTM

PR #682, opened 2026-06-19
MAPA-132 Add overcrowding to establishment roll count

Mjwillis:
## What

Adds overcrowding information to the establishment roll count (`GET /prison/roll-count/{prisonId}` and `.../cells-only/{locationId}`).

A cell is **overcrowded** when it holds more prisoners than its **CNA** (certified normal accommodation). The overcrowding amount is the number of prisoners over the CNA (floored at 0; a null CNA is treated as 0).

## Changes

`ResidentialLocationRollCount` now exposes per-location fields:
- `overcrowded` — boolean; for a cell, whether it is overcrowded; for a parent, whether it contains any overcrowding
- `overcrowdedBy` — for a cell, prisoners over CNA; for a parent, the total across its cells

`LocationRollCount` now exposes rolled-up totals at every level (including the establishment `totals`):
- `cellsOvercrowded` — count of overcrowded cells
- `totalOvercrowded` — total amount of overcrowding

Example: 5 cells each CNA 1, with two cells holding 2 prisoners and one holding 3 → 3 overcrowded cells, total overcrowding 4.

Also populates CNA on the capacity used when building the roll-count tree, so the per-cell calculation has the certified normal accommodation available (this capacity object is internal to the roll-count computation and is not serialized, so no existing output changes).

## Testing

- Unit tests (`PrisonRollCountServiceTest`): worked example, single-cell, at/under-capacity, and null-CNA cases
- Integration tests (`PrisonRollResourceIntTest`): not-overcrowded path asserts the new fields are zero; new test stages 3 prisoners in CNA-2 cells and asserts per-cell and rolled-up overcrowding
- Full suite green (839 tests), ktlint clean

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/ResidentialLocation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/ResidentialLocation.kt
@@ -843,6 +843,7 @@ open class ResidentialLocation(
     capacity = CapacityDto(
       maxCapacity = calcMaxCapacity(),
       workingCapacity = calcWorkingCapacity(),
+      certifiedNormalAccommodation = calcCertifiedNormalAccommodation(),
     ),
     certified = hasCertifiedCells(),
   )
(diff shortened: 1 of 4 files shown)


GurnankCheema (approved):
LGTM

PR #688, opened 2026-06-22
MAPA-138 Reflect local name changes on the current cell certificate

Mjwillis:
## What

For prisons where certification approval is required, changing a location's **local name** does **not** require a fresh certificate approval — but the new name should be reflected on the **current** cell certificate for accuracy. This mirrors the existing behaviour already in place for "used for" types and specialist cell types.

Previously the `PUT /locations/{id}/change-local-name` endpoint updated the location but left the current certificate stale.

## Changes

- **`CellCertificate.kt`** — `localName` on `CellCertificateLocation` is now an `open var` so it can be mutated in place (matching `usedForTypes` / `specialistCellTypes`).
- **`CellCertificateService.kt`** — added `updateLocalNameInCurrentCertificate(location)`, which finds the current certificate for the prison, locates the matching entry by path hierarchy and writes the capitalised local name. Safe no-op when there is no current certificate or the location isn't on it.
- **`LocationService.kt`** — `updateLocalName()` now calls the new method when `activePrisonService.isCertificationApprovalRequired(prisonId)` is true. No approval request is created.
- **`LocationTransformResourceTest.kt`** — new integration test asserting the current certificate reflects the updated local name for an approval-required prison.

## Scope

Hooks the dedicated `change-local-name` endpoint, consistent with how the used-for reflection only hooks its dedicated endpoint. The generic `PATCH /locations/residential/{id}` localName path is out of scope.

## Verification

- `./gradlew build` — passes (tests, ktlint, kover).
- New test `updating local name updates the current cell certificate for the affected location` passes.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/CellCertificate.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/CellCertificate.kt
@@ -140,7 +140,7 @@ open class CellCertificateLocation(
   private val cellMark: String? = null,
 
   @Column(nullable = true)
-  private val localName: String? = null,
+  open var localName: String? = null,
 
   @Column(nullable = false)
   val pathHierarchy: String,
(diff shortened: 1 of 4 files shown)


dafhoustonmoj (approved):
LGTM

PR #689, opened 2026-06-23
MAPA-130 Fix slow cell certificate history by avoiding location over-fetch

Mjwillis:
## Problem

The cell-certificate history page (`…/{prisonId}/cell-certificate/history`) hangs and eventually errors out. It is backed by `GET /cell-certificates/prison/{prisonId}` (the certificate history list).

The list endpoint suffered a JPA over-fetch with a recursive N+1:

- `CellCertificate.locations` was `@OneToMany(fetch = EAGER)` and `CellCertificateLocation.subLocations` is also eager and self-recursive.
- The history query applied the full `cell.certificate.graph` LOAD graph, so for **every historic certificate** Hibernate materialised the **entire prison location tree**, issuing a separate `SELECT` per node down every level.
- `getCellCertificatesForPrison` then maps with `showLocations = false`, so all of that loaded data was **discarded**.

For a large prison (e.g. MDI) with many historic certificates this loaded millions of rows just to render a summary list → the page hung.

## Fix

- Make `CellCertificate.locations` `FetchType.LAZY`.
- Add a lightweight `cell.certificate.summary.graph` that fetches only the one-to-one approval request (avoiding an N+1 across the list) but not the location tree.
- Point `findByPrisonIdOrderByApprovedDateDesc` at the new graph. The by-id and current-certificate endpoints keep the full `cell.certificate.graph` because they genuinely render the tree.

The JSON contract is unchanged — the history list already returned `locations: null`.

## Scope

Certificate history list only. The approval-requests list has a milder version of the same pattern but `Location.childLocations` is already lazy, so it is out of scope here.

## Testing

- Added a regression test asserting the history list returns multiple certificates with summary fields populated and no location tree.
- `./gradlew ktlintCheck build` — green (847 tests).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/CellCertificate.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/CellCertificate.kt
@@ -41,6 +41,15 @@ import java.util.UUID
         ),
       ],
     ),
+    // Lightweight graph for the certificate history list: fetch the one-to-one approval
+    // request (to avoid an N+1 across the list) but NOT the location tree, which the list
+    // does not render (showLocations = false).
+    NamedEntityGraph(
+      name = "cell.certificate.summary.graph",
+      attributeNodes = [
+        NamedAttributeNode("certificationApprovalRequest"),
+      ],
+    ),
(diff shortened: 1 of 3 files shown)


GurnankCheema (approved):
lgtm

PR #690, opened 2026-06-23
MAPB-539 Add cross-cutting VIDEO_ENABLED service type

Mjwillis:
## Why

Prison staff need a "video room calendar" showing every video-capable room in a prison and its free capacity, regardless of which booking service owns the room. Today a room's service type (`VIDEO_LINK`, `OFFICIAL_VISITS`, `APPOINTMENT`, …) is specific to one service, so there is no single way to list "every room that can host a video meeting".

This adds a new **cross-cutting** service type, `VIDEO_ENABLED`, in the existing `VIDEO_LINK_APPOINTMENTS` family. It is a manual label staff add to video rooms (which usually already have another video-related service type). It is purely for visualising capacity — booking still happens in the individual services.

## What

- **`jpa/ServiceUsage.kt`** — added `VIDEO_ENABLED` to `ServiceType` (family `VIDEO_LINK_APPOINTMENTS`, sequence 9) with no NOMIS usage-type or location-type mapping, plus a `crossCutting` helper.
- **`jpa/NonResidentialLocation.kt`** — the NOMIS `sync` path rebuilds a location's service set from NOMIS usage + location-type mappings, which would strip a DPS-only label on the next sync. It now preserves cross-cutting services (mirroring the existing unmapped-usage handling).

Everything else is derived from the enums and needs no change:
- `GET /locations/non-residential/prison/{prisonId}/service/VIDEO_ENABLED` works automatically (`@PathVariable serviceType: ServiceType`).
- `/constants/service-types` and `/constants/service-family-types` expose the new value automatically (the UI's service checkboxes are driven from these).
- No DB migration needed (`service_usage.service_type` is `varchar(80)`, stored as string). No backfill — staff add the label manually.

## Tests

- Updated `LocationConstantsIntTest` expectations for both constants endpoints.
- Added an endpoint test for `/service/VIDEO_ENABLED`.
- Added a sync-preservation test confirming a NOMIS re-sync keeps `VIDEO_ENABLED`.
- `./gradlew build` passes (full suite + ktlint).

## Follow-up (UI)

No production code change in the non-residential UI — the service checkboxes render from the constants endpoint, so `VIDEO_ENABLED` appears under "Book a video link" once this is on dev. The UI's generated API types should be refreshed (`npm run generate-location-api-types`) after deployment.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/NonResidentialLocation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/NonResidentialLocation.kt
@@ -254,6 +254,9 @@ class NonResidentialLocation(
       }
     }
 
+    // Preserve DPS-only cross-cutting services NOMIS does not manage (e.g. VIDEO_ENABLED)
+    serviceTypeForLocation.addAll(services.map { it.serviceType }.filter { it.crossCutting })
+
     updateServices(serviceTypeForLocation, upsert.lastUpdatedBy, clock, linkedTransaction)
 
     upsert.internalMovementAllowed?.let { internalMovementAllowedUpdate ->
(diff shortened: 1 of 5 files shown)


timharrison-moj (approved):
LGTM. From a limited understanding of locations API generally.

PR #696, opened 2026-06-26
MAPA-149 Move approval type descriptions to backend

Mjwillis:
## What

Moves approval type descriptions from the frontend into the backend, where the `ApprovalType` enum is the source of truth, and exposes them through the existing constants endpoint pattern. Also applies content-team wording changes.

## Changes

- **`ApprovalType` enum** gains a `description` for all 13 values.
- **New `GET /constants/approval-type` endpoint** (mirrors `/constants/deactivated-reason`, guarded by `ROLE_READ_LOCATION_REFERENCE_DATA`).
- **Content changes** in `CellCertificateUploadProcessingService`:
  - Ingestion failures now report `Location not found on Residential locations`.
  - Upload-generated approval requests record a fixed explanation: _"This is the cell certificate that was imported when the prison started using Residential locations to manage its cell certificate."_
- **New/changed descriptions:** `PERMANENT_DEACTIVATION` → "Archived", `PRISON_BASELINE` → "Initial certificate generation", `CELL_CERTIFICATE_UPLOAD` → "Initial cell certificate import".

## Tests

- New integration test for `GET /constants/approval-type` asserting all 13 key/description pairs.
- Updated the cell certificate ingestion failure-message assertion.
- `LocationConstantsIntTest` + `CellCertificateUploadProcessingIntTest` pass; `ktlintCheck` clean.

## Notes

The companion frontend change (consuming this endpoint) is on a separate branch/PR under the same ticket. This backend should be deployed first; the frontend degrades gracefully to the raw enum key until the endpoint is available.


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/approvalrequest/CertificationApprovalRequest.kt
```diff
@@ -168,19 +168,20 @@ enum class ApprovalRequestStatus {
...
-  CELL_CERTIFICATE_UPLOAD,
+  SIGNED_OP_CAP("Change signed operational capacity"),
+  DRAFT("Add new locations to certificate", true),
+  DEACTIVATION("Deactivation"),
+  PERMANENT_DEACTIVATION("Archived"),
+  CELL_MARK("Change cell door number", true),
+  CELL_SANITATION("Change cell sanitation", true),
+  REACTIVATION("Activation"),
+  CAPACITY_CHANGE("Cell capacity", true),
+  SPECIALIST_CELL_TYPE("Set special cell type", true),
```
jimbali, line 182: I'm not sure how practical this is going to be, because some of these have to be determined dynamically in the UI, for example `SPECIALIST_CELL_TYPE` could be "Set special cell type" or "Remove special cell type" depending on whether you are removing or adding the type. Or are you intending to implement that logic in the API too?

jimbali: See: https://github.com/ministryofjustice/hmpps-locations-inside-prison/blob/main/server/formatters/approvalTypeDescription.ts

Mjwillis: Good question — no, the dynamic wording stays in the UI. DEACTIVATION, REACTIVATION and SPECIALIST_CELL_TYPE keep their existing branches in approvalTypeDescription.ts (see the companion frontend PR), so the location-type-based phrasing and the Set/Remove prefix are unchanged. The API description is only used for the genuinely static types; for the dynamic ones it's just a canonical default the formatter overrides. The formatter also falls back to the raw enum key, so nothing hard-depends on the constant.

Mjwillis: See https://github.com/ministryofjustice/hmpps-locations-inside-prison/pull/679

jimbali (approved):

PR #698, opened 2026-06-26
MAPB-490 Add batch non-residential location lookup by ids

Mjwillis:
## What
New `POST /locations/non-residential/batch` — takes a JSON array of location ids and returns the matching non-residential locations (`NonResidentialLocationDTO`) in one call.

## Why
The prisoner-property-api needs to resolve many property-storage location names when listing a prisoner's property. Today that's one `GET /locations/{id}` per id; this batch endpoint makes it a single round-trip.

## How
- `NonResidentialService.getByIds(ids)` uses the non-residential repository's inherited `findAllById`, so only non-residential ids resolve — residential or unknown ids are silently omitted (no 404).
- Reuses the existing `NonResidentialLocationDTO` / `toNonResidentialDto()` mapper, consistent with the other non-resi endpoints.
- `@PreAuthorize("hasRole('ROLE_VIEW_LOCATIONS')")`, full springdoc annotations, modelled on the existing `POST /locations/keys` batch.
- Tests in `LocationNonResidentialResourceTest`: security (401/403), happy path, and filtering (residential + unknown ids excluded).

## Notes
- Consumers calling this (e.g. the property-api client credentials) need the `ROLE_VIEW_LOCATIONS` role granted — a config concern, not code here.
- Repointing the property-api's `LocationsClient.getLocations` at this endpoint is a follow-up in that repo.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationNonResidentialResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationNonResidentialResource.kt
@@ -90,6 +90,41 @@ class LocationNonResidentialResource(
   ) = nonResidentialService.getById(id = id)
     ?: throw LocationNotFoundException(id.toString())
 
+  @PostMapping("/non-residential/batch", produces = [MediaType.APPLICATION_JSON_VALUE])
+  @ResponseStatus(HttpStatus.OK)
+  @PreAuthorize("hasRole('ROLE_VIEW_LOCATIONS')")
+  @Operation(
+    summary = "Returns non-residential location information for the supplied ids in one call",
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/LocationNonResidentialResource.kt
```diff
@@ -90,6 +90,41 @@ class LocationNonResidentialResource(
   ) = nonResidentialService.getById(id = id)
     ?: throw LocationNotFoundException(id.toString())
 
+  @PostMapping("/non-residential/batch", produces = [MediaType.APPLICATION_JSON_VALUE])
```
jimbali, line 93: I guess it's a POST rather than a GET so that we can pass in an enormous list of IDs in the POST body?

jimbali (approved):
lgtm

PR #715, opened 2026-07-13
MAPB-651 Add location hierarchy to non-residential location DTO

Mjwillis:
## What

Adds a `locationHierarchy` field to `NonResidentialLocationDTO` (ancestors **and** self, ordered top → bottom), populated from the existing `getHierarchy()` helper — mirroring how `ResidentialSummary` already exposes `locationHierarchy`.

It flows through the shared `NonResidentialLocation.toNonResidentialDto()` mapper, so all three read endpoints get it from one place:
- `GET /locations/non-residential/summary/{prisonId}`
- `GET /locations/non-residential/{id}`
- batch get-by-ids

Ancestor local names are passed through `formatLocation(...)` so they render title-cased, consistent with the DTO's own `localName` (rather than the raw upper-cased value `getHierarchy()` returns).

## Why

In the non-residential locations UI, parents and their children appear in a flat list and are often **named identically** (e.g. a parent "Gym" and a child "Gym"), so users can't tell them apart or decide which to archive. This field lets the UI show the readable parent/hierarchy path for each location.

## Testing

- Extended the summary happy-path assertions in `LocationNonResidentialResourceTest` to verify `locationHierarchy` (`[wingZ, adjRoom]`, with formatted names/levels).
- Full `LocationNonResidentialResourceTest` (84 tests) and `NonResidentialServiceTest` pass; ktlint clean.

## Related

Paired with the UI change in `hmpps-non-residential-locations-ui` (MAPB-651). **This PR should merge and deploy to dev first** so the field is populated before the UI consumes it.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/NonResidentialLocation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/jpa/NonResidentialLocation.kt
@@ -125,6 +125,7 @@ class NonResidentialLocation(
       isLeafLevel = isLeafLevel(),
       prisonId = prisonId,
       parentId = getParent()?.id,
+      locationHierarchy = getHierarchy().map { it.copy(localName = it.localName?.let(::formatLocation)) },
       level = getLevel(),
       permanentlyInactive = isPermanentlyDeactivated(),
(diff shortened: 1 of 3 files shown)


dafhoustonmoj (approved):
LGTM

danbenton-mojdt (approved):

PR #737, opened 2026-08-21
MAPA-262: Prevent error when new emails collide with old ones

jimbali:
## Fix: duplicate key error when replacing notification mailbox with an overlapping address

`PUT /prison-configuration/{prisonId}/notification-mailboxes/{notificationGroup}` returned a 500 when the new list of email addresses overlapped with the existing ones (e.g. re-submitting the same address, or adding one email to an existing list).

### Root cause

`replaceMailboxes` deletes the existing rows for the prison + group, then inserts the new set, within a single transaction. Hibernate's flush always executes queued **inserts before deletes**, regardless of call order — so the new `INSERT`s ran before the old rows were actually removed, tripping the `prison_notification_mailbox_unique_idx` constraint whenever an address appeared in both the old and new lists.

### Fix

Call `prisonNotificationMailboxRepository.flush()` immediately after the delete, forcing the removal to hit the database before the new rows are inserted.

### Testing

- Added an integration test (`can replace notification mailbox when new list overlaps with existing addresses`) reproducing the reported scenario
- Full test suite passes (973 tests)


df-oc (approved):
LGTM

PR #739, opened 2026-08-24
MAPA-262: Allow for all-prisons mailbox

jimbali:
## Add default notification mailboxes

Adds support for global/default notification mailboxes alongside prison-specific mailbox configuration.

### What changed

- Added support for default mailbox rows using `prison_id = null`
- Added default mailbox endpoints:
  - `GET /prison-configuration/notification-mailboxes/defaults/{notificationGroup}`
  - `PUT /prison-configuration/notification-mailboxes/defaults/{notificationGroup}`
  - `DELETE /prison-configuration/notification-mailboxes/defaults/{notificationGroup}`
- Updated prison-specific mailbox lookup to fall back to the default mailbox when no prison-specific rows exist
- Added `source` to mailbox responses to indicate whether the result came from `PRISON` or `DEFAULT`
- Added a migration to allow nullable `prison_id` and separate unique constraints for prison-specific vs default mailboxes
- Kept audit, telemetry, and log entries redacted so email addresses are not written to those sinks

### Testing

- Added repository coverage for default mailbox rows and uniqueness
- Added service coverage for default fallback and default replacement
- Added integration coverage for default mailbox endpoints and fallback behaviour
- Full test suite passes

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/PrisonNotificationMailboxService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/PrisonNotificationMailboxService.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.locationsinsideprison.service
 
+import com.fasterxml.jackson.annotation.JsonInclude
 import com.microsoft.applicationinsights.TelemetryClient
 import io.swagger.v3.oas.annotations.media.Schema
 import jakarta.validation.ValidationException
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/PrisonNotificationMailboxService.kt
```diff
@@ -33,29 +34,76 @@ class PrisonNotificationMailboxService(
...
+    if (mailboxes.isNotEmpty()) {
+      return mailboxes.toDto(prisonId, notificationGroup, NotificationMailboxSource.PRISON)
+    }
+
+    val defaultMailboxes = prisonNotificationMailboxRepository.findByPrisonIdIsNullAndNotificationGroup(notificationGroup)
+    if (defaultMailboxes.isEmpty()) {
       throw PrisonNotificationMailboxNotFoundException(prisonId, notificationGroup)
     }
-    return mailboxes.toDto(prisonId, notificationGroup)
+    return defaultMailboxes.toDto(prisonId, notificationGroup, NotificationMailboxSource.DEFAULT)
```
Mjwillis, line 57: **Prison-level delete becomes unexpressible / silently reverts to the default.**

With this fallback, `GET /prison-configuration/LEI/notification-mailboxes/CERT_ADMIN` now returns 200 (`source=DEFAULT`) when only defaults exist — but `deleteMailboxes` (line 155) still hard-404s for that same prison+group. So an admin UI renders the addresses, the user clicks delete, and gets a 404 for a resource they were just shown.

In the other direction: after successfully deleting LEI's own rows, the very next GET returns the default addresses again. There is no way to express "do not notify this prison", and an offboarded prison-specific mailbox silently falls back to the global one.

Worth deciding deliberately (a tombstone row, or `?includeDefault=false`) rather than inheriting it by accident.

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/PrisonNotificationMailboxService.kt
```diff
@@ -33,29 +34,76 @@ class PrisonNotificationMailboxService(
...
+  }
+
+  fun getDefaultMailboxes(notificationGroup: NotificationGroup): PrisonNotificationMailboxDto {
+    val mailboxes = prisonNotificationMailboxRepository.findByPrisonIdIsNullAndNotificationGroup(notificationGroup)
+    if (mailboxes.isEmpty()) {
+      throw PrisonNotificationMailboxNotFoundException("defaults", notificationGroup)
```
Mjwillis: **Misleading 404 body.** `PrisonNotificationMailboxNotFoundException("defaults", ...)` renders as:

> There is no notification mailbox found for prisonId = defaults, notificationGroup = ...

which surfaces a prison code that doesn't exist to API clients. Same at line 176 in `deleteDefaultMailboxes`. A dedicated exception — or making `prisonId` nullable on the existing one — would produce an accurate message.

src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/service/PrisonNotificationMailboxService.kt
```diff
@@ -33,29 +34,76 @@ class PrisonNotificationMailboxService(
 
   companion object {
     val log: Logger = LoggerFactory.getLogger(PrisonNotificationMailboxService::class.java)
+    private const val DEFAULT_MAILBOX_AUDIT_PRISON_ID = "ALL"
```
Mjwillis: **`"ALL"` gets written into `linked_transaction.prison_id`.** That column holds a real prison code everywhere else in the schema. There's no FK so it won't fail, but any audit report grouping or filtering transactions by prison will now surface a phantom prison "ALL", indistinguishable from a genuine (if invalid) code.

Consider a value that can't be mistaken for a prison code, or recording the scope in `transactionDetail` only and leaving `prisonId` out.

src/test/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/resource/PrisonNotificationMailboxResourceTest.kt
```diff
@@ -245,6 +277,68 @@ class PrisonNotificationMailboxResourceTest : SqsIntegrationTestBase() {
     }
   }
 
+  @DisplayName("DELETE /prison-configuration/{prisonId}/notification-mailboxes/{notificationGroup}")
+  @Nested
+  inner class DefaultNotificationMailboxTest {
```
Mjwillis: Two things on this new nested class:

1. **Wrong `@DisplayName`** — it's a copy of the one on `DeleteNotificationMailboxTest` (line 342), so test reports will show two identically named sections and neither describes the defaults endpoints.
2. **No `Security` block** — unlike every other nested class in this file. The three new `ROLE_LOCATION_CONFIG_ADMIN`-gated endpoints have zero 401/403 coverage, so a dropped or mistyped `@PreAuthorize` on an admin-only mutating endpoint would pass CI.

Mjwillis (commented):
A few review notes, posted inline.

Mjwillis (commented):
Some minor comments

Mjwillis (approved):
LGTM

